PR #221, opened 2019-06-13
Enhance logs when panic.

ldez:


spatecon:
And how to disable these logs if panics is a part of execution?

Issue #280, opened 2019-07-25
Entering 1=1 triggers a panic


maxhawkins:
OS: macOS 10.14.4
Commit: 2dfede3b90b40723cb2c015adb2f2ab5227825cd

Using the repl:

```sh
~> yaegi
> 1=1    
1:28: panic
panic: reflect: reflect.Value.Set using unaddressable value [recovered]
	panic: reflect: reflect.Value.Set using unaddressable value

goroutine 1 [running]:
github.com/containous/yaegi/interp.runCfg.func1(0xc00018a050, 0xc000112a00)
	/Users/max/go/src/github.com/containous/yaegi/interp/run.go:107 +0x1e9
panic(0x17fc540, 0xc000180250)
	/usr/local/Cellar/go/1.12.1/libexec/src/runtime/panic.go:522 +0x1b5
reflect.flag.mustBeAssignable(0x82)
	/usr/local/Cellar/go/1.12.1/libexec/src/reflect/value.go:234 +0x13a
reflect.Value.Set(0x17fb400, 0xc0000ee650, 0x82, 0x17fb400, 0xc0000ee660, 0x82)
	/usr/local/Cellar/go/1.12.1/libexec/src/reflect/value.go:1467 +0x2f
github.com/containous/yaegi/interp.assign.func3(0xc00018a050, 0x1945db0)
	/Users/max/go/src/github.com/containous/yaegi/interp/run.go:256 +0xb6
github.com/containous/yaegi/interp.runCfg(0xc000112a00, 0xc00018a050)
	/Users/max/go/src/github.com/containous/yaegi/interp/run.go:112 +0x70
github.com/containous/yaegi/interp.(*Interpreter).run(0xc0000ce180, 0xc000112900, 0x0)
	/Users/max/go/src/github.com/containous/yaegi/interp/run.go:93 +0x126
github.com/containous/yaegi/interp.(*Interpreter).Eval(0xc0000ce180, 0xc0000ee5f0, 0x6, 0xc0001c9df0, 0x5, 0x19136b5, 0x1, 0xc0000ee5f0)
	/Users/max/go/src/github.com/containous/yaegi/interp/interp.go:277 +0x2f9
github.com/containous/yaegi/interp.(*Interpreter).Repl(0xc0000ce180, 0xc000010010, 0xc0000d0000)
	/Users/max/go/src/github.com/containous/yaegi/interp/interp.go:320 +0x1c3
main.main()
[...]

PR #297, opened 2019-07-27
feat: improve panic management in REPL

bogen85:
Entering 1=1 triggers a panic #280
Added TrapEval as an Eval wrapper, to trap a panic and return the prompt.
Trap only occurs when running interactively in a Terminal session.
When not running interactively panic is not trapped.


interp/interp.go
```diff
@@ -335,9 +336,29 @@ func (interp *Interpreter) Repl(in, out *os.File) {
 	}
 }
 
+// wrapper around Eval that traps a panic if the input is a terminal
+func (interp *Interpreter) TrapEval(in *os.File, src string) (reflect.Value, error) {
+	if runningInTerminal(in) {
```
ldez: Can be simplify like that:
```go
if in == os.Stdin {
//...
}
```

bogen85: Alright, I'll un-export. I did not see your reply here. I was using the same check that was already present. (for the prompt). I just combined the check into on one function....

ldez: yes it's not directly related to your changes, it's something that I have missing previously.

ldez: you can keep like that but you can simplify the function:
```go
func runningInTerminal(in *os.File) bool {
	stat, err := in.Stat()
	return err == nil && stat.Mode()&os.ModeCharDevice != 0
}
```

bogen85: I just tried it with
```go
if in == os.Stdin {
//...
}
```
That causes yaegi to keep running after a panic when stdin is redirected from a file.

bogen85: Alright, so two PRs now, one with TrapEval and one with trapEval and both are passing.

bogen85: Maybe the check should only be done once. (for both the prompt and the eval)
Unless there is some common condition where stdin starts out as Terminal input, but then changes to non Terminal input.

ldez: ok good catch.

could you just simplify the function?

```go
func runningInTerminal(in *os.File) bool {
	stat, err := in.Stat()
	return err == nil && stat.Mode()&os.ModeCharDevice != 0
}
``

Comment (bogen85) [reply interp/interp.go]: Oh, the prompt check was only once. I'm making the trapEval check it only once as well.

**Comment by ldez** [comment]
FYI you don't need to keep the PR title in your commits and you also don't need to add the PR number.

Github allows to squash the commits before the merge and sets the good commit message based on the PR title.

File: interp/interp.go [hunk 2]
```diff
@@ -335,9 +337,27 @@ func (interp *Interpreter) Repl(in, out *os.File) {
 	}
 }
 
+// trapEval is a wrapper around Eval that traps a panic if the input is a terminal
+func (interp *Interpreter) trapEval(inTerminal bool, src string) (reflect.Value, error) {
```
ldez: @mvertes I'm not sure about the naming `trapEval` (seems too "bash" for me) WDYT?

bogen85: Yeah, on the lint part...

I did not make any change to yaegi.go
And in get prompt it was already "out" for the second parameter.

As far as "trapEval"...

recoveredEval?
checkedEval?
userEval?

```
$ golangci-lint run
cmd/yaegi/yaegi.go:1: : 
cmd/yaegi/test.go:1:1: expected 'package', found 1 (typecheck)
/*
cmd/yaegi/yaegi.go:32:2: could not import flag (no metadata for flag) (typecheck)
	"flag"
	^
cmd/yaegi/yaegi.go:33:2: could not import fmt (no metadata for fmt) (typecheck)
	"fmt"
	^
interp/interp.go:359:33: `out` can be `io.Writer` (interfacer)
func getPrompt(inTerminal bool, out *os.File) func() {
```

interp/interp.go
```diff
@@ -335,9 +337,27 @@ func (interp *Interpreter) Repl(in, out *os.File) {
...
+// runningInTerminal returns true only if input is a terminal
+func runningInTerminal(in *os.File) bool {
+	stat, err := in.Stat()
+	return err == nil && stat.Mode()&os.ModeCharDevice != 0
+}
+
 // getPrompt returns a function which prints a prompt only if input is a terminal
-func getPrompt(in, out *os.File) func() {
-	if stat, err := in.Stat(); err == nil && stat.Mode()&os.ModeCharDevice != 0 {
+func getPrompt(inTerminal bool, out *os.File) func() {
```
ldez: from the linter:
```console
$ golangci-lint run
interp/interp.go:359:33: `out` can be `io.Writer` (interfacer)
func getPrompt(inTerminal bool, out *os.File) func() {
```

You have to replace the type of the function parameter by `io.Writer`. Only in the function signature.

interp/interp.go
bogen85: After reviewing Entering 1=1 triggers a panic #280 some more I don't really believe this PR will really truly fix the underlying problem. It does make it more manageable, but one underlying issue is the panics are not recovered for correctly. Something else gets messed up. See my further comments on #280

ldez: yes your PR don't fix #280, as you say: make the panic more manageable.

interp/interp.go
mvertes: After looking at a broader scope, I think it may be not necessary to fix the REPL, at least not in this order. We have to perform error processing correctly first (in the core of interpreter), so the only panics which occur are the ones we want to propagate anyway, even in REPL mode. Sorry for the wasted time.

ldez: it seems interesting to me as a first step before treating the problem at the root.
This is not enough and we will change the behavior in the future, but it can give an impression of "robustness"

mvertes: I will avoid to respond too fast, and make another mistake 😅

ldez: another point: this allow to write `panic` in the REPL without that the command-line panic also.

bogen85:
Per our earlier discussion (in this PR) I'm closing this PR.

Issue #304, opened 2019-07-29
crash regardless of arguments


robpike:
I'm very excited about this project but all I can get is a panic immediately on startup. I installed with `go get -u github.com/containous/yaegi/cmd/yaegi`.

```
% ls p.go
p.go
% yaegi p.go
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/containous/yaegi/interp.goNumVersion(0x2)
	/Users/r/src/github.com/containous/yaegi/interp/build.go:91 +0xa5
github.com/containous/yaegi/interp.init()
	/Users/r/src/github.com/containous/yaegi/interp/build.go:60 +0x26
% yaegi
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/containous/yaegi/interp.goNumVersion(0xc00009a1c0)
	/Users/r/src/github.com/containous/yaegi/interp/build.go:91 +0xa5
github.com/containous/yaegi/interp.init()
	/Users/r/src/github.com/containous/yaegi/interp/build.go:60 +0x26
% yaegi -help
panic: runtime error: index out of range [1] with length 1

goroutine 1 [running]:
github.com/containous/yaegi/interp.goNumVersion(0xc0000a61b0)
	/Users/r/src/github.com/containous/yaegi/interp/build.go:91 +0xa5
github.com/containous/yaegi/interp.init()
	/Users/r/src/github.com/containous/yaegi/interp/build.go:60 +0x26
% go version
go version devel +54c6b814ac Tue Jul 2 20:07:52 2019 +0000 darwin/amd64
% uname -a
Darwin bismarck.local 18.7.0 Darwin Kernel Version 18.7.0: Thu Jun 20 18:42:21 PDT 2019; root:xnu-4903.270.47~4/RELEASE_X86_64 x86_64
% 
```

ldez:
Hello,

Yaegi only works with go1.11 and go1.12

Duplicate of #285

robpike:
Well, that's easy. Thanks for the quick reply.

ldez:
We can allow the use of any version of Go, if we change the way of handling the build tags and if you provide manually the stdlib mapping (by using our `goexports` tool)

robpike:
I am happy to wait for 1.13 support. I got it running with 1.12 but the first program I wanted to use it with has some 1.13 features, so it's clear I should just wait.

PR #306, opened 2019-07-30
interp/build: improve Go version handling.

ldez:
Display a better message when a version is not supported.

Allow to use `devel`, `beta`, and `rc` versions.

Related to #304, #309

### DIFF
--- a/interp/build.go
+++ b/interp/build.go
@@ -2,6 +2,7 @@ package interp
 
 import (
 	"go/parser"
+	"math"
 	"path"
 	"runtime"
 	"strconv"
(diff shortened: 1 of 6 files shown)


interp/build.go
```diff
@@ -86,9 +87,29 @@ func buildTagOk(s string) (r bool) {
 }
 
 // goNumVersion returns the go minor version number
-func goNumVersion() int {
-	v := strings.Split(runtime.Version(), ".")
-	n, _ := strconv.Atoi(v[1])
+func goNumVersion(version string) int {
```
sbinet: any reason why we don't use `go/build.Context` instead of `runtime.Version`?
- https://godoc.org/go/build#Context

ie: a combination of `build.Default.ReleaseTags` and `build.Default.BuildTags` ?

with my current dev environment:
```
$> go version
go version devel +60f14fddfe Wed Jun 26 16:35:14 2019 +0000 linux/amd64
```

the following program:
```go
package main

import (
	"go/build"
	"log"
)

func main() {
	ctx := build.Default
	log.Printf("build-tags:   %v", ctx.BuildTags)
	log.Printf("release-tags: %v", ctx.ReleaseTags)
}
```

prints:
```
$> go run ./main.go
2019/07/30 10:38:50 build-tags:   []
2019/07/30 10:38:50 release-tags: [go1.1 go1.2 go1.3 go1.4 go1.5 go1.6 go1.7 go1.8 go1.9 go1.10 go1.11 go1.12 go1.13]
```

ldez: The `build.Default.ReleaseTags` simplify the version parsing for devel/beta/rc.

`build.Default.BuildTags` seems less relevant for handling version.
 
> BuildTags is not set for the Default build Context.

It can be useful for handling custom build tags, for that we could to add a `build.Context` to the `buildOk` method and allow to define custom build context as an option of the interpreter.

The purpose of this PR is to manage the version, then I will do another PR after the merge of this PR to manage that.

PR #311, opened 2019-07-30
cmd/goexports: add -license flag to goexports

### LINKED ISSUE #310: cmd/goexports: support LICENSE blurb
in some projects (_e.g._ Apache Arrow), all files are required to have a LICENSE blurb at the start (even automatically generated code).

one should consider adding a `-license ./LICENSE.txt` flag to `goexports` to automatically add it.

sbinet:
Fixes #310.


sbinet:
PTAL

cmd/goexports/goexports.go
```diff
@@ -53,6 +57,7 @@ import (
 
 func init() {
 	Symbols["{{.PkgName}}"] = map[string]reflect.Value{
+		{{- if len .Val | ne 0}}
```
ldez: This check seems not needed because the range syntax already support empty map.

cmd/goexports/goexports.go
```diff
@@ -61,16 +66,21 @@ func init() {
 				"{{$key}}": reflect.ValueOf({{$value.Name}}),
 			{{end -}}
 		{{end}}
+		{{- end}}
 
+		{{- if len .Typ | ne 0}}
```
ldez: This check seems not needed because the range syntax already support empty map.

cmd/goexports/goexports.go
```diff
@@ -61,16 +66,21 @@ func init() {
...
 		// type definitions
 		{{range $key, $value := .Typ -}}
 			"{{$key}}": reflect.ValueOf((*{{$value}})(nil)),
 		{{end}}
+		{{- end}}
 
+		{{- if len .Wrap | ne 0}}
```
ldez: This check seems not needed because the range syntax already support empty map.

cmd/goexports/goexports.go
sbinet: yes, that was the goal.

cmd/goexports/goexports.go
```diff
@@ -258,15 +269,59 @@ func fixConst(name string, val constant.Value) string {
...
+	license := new(strings.Builder)
+	sc := bufio.NewScanner(bytes.NewReader(raw))
+	for sc.Scan() {
+		txt := sc.Text()
+		switch txt {
+		case "":
+			license.WriteString("//\n")
+		default:
+			license.WriteString("// " + txt + "\n")
+		}
```
ldez: can be simplify:
```go
	for sc.Scan() {
		txt := sc.Text()
		if txt != "" {
			txt = " " + txt
		}
		license.WriteString("//" + txt + "\n")
	}
```

cmd/goexports/goexports.go
```diff
@@ -258,15 +269,59 @@ func fixConst(name string, val constant.Value) string {
...
+		return "", nil
+	}
+
+	raw, err := ioutil.ReadFile(fname)
+	if err != nil {
+		return "", fmt.Errorf("could not read LICENSE file content: %v", err)
+	}
+
+	license := new(strings.Builder)
+	sc := bufio.NewScanner(bytes.NewReader(raw))
```
ldez: as the file content is only used in the scanner, it's possible to simplify:
```go
	file, err := os.Open(fname)
	if err != nil {
		return "", fmt.Errorf("could not open LICENSE file: %v", err)
	}
	defer file.Close()

	license := new(strings.Builder)
	sc := bufio.NewScanner(file)
```

cmd/goexports/goexports.go
```diff
@@ -212,7 +222,7 @@ func genContent(dest, pkgName string) ([]byte, error) {
 		buildTags += ",!windows,!nacl,!plan9"
 	}
 
-	b := &bytes.Buffer{}
+	b := new(bytes.Buffer)
```
ldez: could you explain this change?

cmd/goexports/goexports.go
kortschak: I would place these conditions (`.Typ` and `.Wrap`) one line up. At the moment they still render the blank line if the map is empty, leaving potentially 2 blank lines in the case where there is no type or interface symbol to write.

cmd/goexports/goexports.go
sbinet: it's usually spelled this way in Go's stdlib.

cmd/goexports/goexports.go
sbinet: to handle this case, one would need to blank-import the package being wrapped.
(I'll add that. I stumbled on this for the _e.g._ `runtime/race` package.)

sbinet:
travis failed because of:
https://travis-ci.com/containous/yaegi/jobs/221238375#L284

```
golangci-lint run
cmd/goexports/goexports.go:282:15: Error return value of `f.Close` is not checked (errcheck)
	defer f.Close()
	             ^
Makefile:3: recipe for target 'check' failed
```

seems a bit too strict a check... (if you ask me)

what is the expected yaegi-way to silence/fix those?

cmd/goexports/goexports.go
```diff
@@ -258,15 +268,58 @@ func fixConst(name string, val constant.Value) string {
...
+func genLicense(fname string) (string, error) {
+	if fname == "" {
+		return "", nil
+	}
+
+	f, err := os.Open(fname)
+	if err != nil {
+		return "", fmt.Errorf("could not open LICENSE file: %v", err)
+	}
+	defer f.Close()
```
ldez:
```suggestion
	defer func() { _ := f.Close()}
```

sbinet:
PTAL

cmd/goexports/goexports.go
```diff
@@ -61,16 +66,19 @@ func init() {
 				"{{$key}}": reflect.ValueOf({{$value.Name}}),
 			{{end -}}
 		{{end}}
-
+		{{- end}}
+		{{- if .Typ}}
```
kortschak: If you have a blank like below this and the `.Wrap` conditional below, you'll get nice stanza formatting for these groups of symbols.

sbinet: actually did the reverse (put the blank line at the end of the block) to better handle the case where, _e.g._ there would be no `.Val` but `.Typ`. (following your recommandation, a blank line between `map[string]reflect.Value{` and `// type definitions` would have been inserted.)

sbinet:
PTAL

cmd/goexports/goexports.go
kortschak: This is still not right; if you have `.Val` or `.Typ` but no `.Wrap` you end up with a trailing blank.

The logic, when not done in a loop is actually sort of convoluted. Maybe not caring is better.

sbinet: the file is `gofmt`'d so the trailing blank line should be removed (I think)

cmd/goexports/goexports.go
```diff
@@ -212,15 +222,17 @@ func genContent(dest, pkgName string) ([]byte, error) {
...
-		"BuildTags": buildTags,
+		"Dest":        dest,
+		"Imports":     imports,
+		"PkgName":     pkgName,
+		"Val":         val,
+		"Typ":         typ,
+		"Wrap":        wrap,
+		"BuildTags":   buildTags,
+		"License":     license,
+		"BlankImport": len(val)+len(typ)+len(wrap) == 0,
```
ldez: It's not related to the license support, so could you remove those changes from this PR, thanks.

sbinet: done

sbinet:
PTAL

(anything I overlooked before this could be merged so I could tackle the "blank import" thing and the speed improvement of `goexports` proper?)

PR #318, opened 2019-07-30
interp/build: support custom build constraints.

### LINKED ISSUE #307: [question]: building from source based on build tags
This follows on from #301.

If we have source that provides both a pure Go implementation and an asm optimisation, like for example the "math" package, then importing the package will presumably need to only build the pure Go code. By default, packages that do this tend to provide the asm implementation, falling back to pure Go for unsupported GOARCHs and sometime allowing build tags to require pure Go (this is the case for "gonum.org/v1/gonum/mat", where the "noasm" tag build only pure Go Gonum packages).

Is there a way to allow this information to be passed to the interpreter?

ldez:
Use of build.Context to handle custom build tags. Let user define a list of build tags in interpreter options.

Fixes #307

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -3,6 +3,7 @@ package interp
 import (
 	"bufio"
 	"fmt"
+	"go/build"
 	"go/scanner"
 	"go/token"
 	"os"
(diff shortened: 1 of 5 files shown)


interp/interp.go
```diff
@@ -58,10 +59,11 @@ type Exports map[string]map[string]reflect.Value
...
 type opt struct {
-	astDot bool   // display AST graph (debug)
-	cfgDot bool   // display CFG graph (debug)
-	noRun  bool   // compile, but do not run
-	goPath string // custom GOPATH
+	astDot  bool   // display AST graph (debug)
+	cfgDot  bool   // display CFG graph (debug)
+	noRun   bool   // compile, but do not run
+	goPath  string // custom GOPATH
+	context build.Context
```
mvertes: Please add a comment for this field like: `// To handle build constraints, and skip unwanted source files`

PR #324, opened 2019-08-01
Updated alias for installation from readme.md

mihaicc:
go get was not enough for me to get in a position to simpli write `yaegi` and work.
I'm on macos mojave with golang 1.12

Update the doc with how i fixed it. If it works with the previous directions on linux, i'd propose a mix of the two in the readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,7 @@ go get -u github.com/containous/yaegi/cmd/yaegi
 ```
 
 Note that you can use [rlwrap](https://github.com/hanslub42/rlwrap) (install with your favorite package manager),
-and alias the `yaegi` command in `alias yaegi='rlwrap yaegi'` in your `~/.bashrc`, to have history and command line edition.
+and alias the `yaegi` command in `alias yaegi='rlwrap /Users/your-username/your-workspace-dir/bin/yaegi'` in your `~/.bash_profile`, to have history and command line edition.
 
 ## Usage
 


README.md
```diff
@@ -35,7 +35,7 @@ go get -u github.com/containous/yaegi/cmd/yaegi
 ```
 
 Note that you can use [rlwrap](https://github.com/hanslub42/rlwrap) (install with your favorite package manager),
-and alias the `yaegi` command in `alias yaegi='rlwrap yaegi'` in your `~/.bashrc`, to have history and command line edition.
+and alias the `yaegi` command in `alias yaegi='rlwrap /Users/your-username/your-workspace-dir/bin/yaegi'` in your `~/.bash_profile`, to have history and command line edition.
```
theclapp: There are a couple problems with this.

* `/Users` is macOS-specific.
* `~/.bashrc` is actually the better place for aliases than `~/.bash_profile`.
* Not everyone uses bash.  e.g. I use zsh.
* "command line editing" not "edition" (which was there already).

You can't use $HOME, either (instead of `/Users/...`) — you can't assume everyone has their workspace under $HOME. (Lots of people do, but it's not _required_.) You can't use $GOPATH as-is — $GOPATH can have multiple paths in it, just like $PATH. (And $GOPATH is deprecated or will be soon anyway, because modules.)

Should probably be something more like

> and alias the `yaegi` command in `alias yaegi='rlwrap yaegi'` in your shell's rc file (e.g. `~/.bashrc` or `~/.zshrc`), to have history and command line editing. You may need to supply a full path to the `yaegi` binary, depending on how you have your $PATH set up.

mvertes:
Thank you for your proposal. For simplicity and clarity, we prefer to keep the current wording as is for now. There may be variations according to different user settings, we just chose one.

Issue #257, opened 2019-07-24
importing "github.com/spf13/cobra" panics


fiws:
The yaegi repl panics when i try to import "github.com/spf13/cobra"

Not sure if i am doing anything wrong, as yaegi does not seem to find any other libs i throw at it. (always getting "unable to find source related to). But it should not panic.

Expected result:
```console
 yaegi                        
> import "github.com/spf13/cobra"

```

Got:
```console
 yaegi                        
> import "github.com/spf13/cobra"
panic: reflect: FieldByName of non-struct type

goroutine 1 [running]:
reflect.(*rtype).FieldByName(0xc000097da0, 0xc0002c61db, 0x5, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/usr/lib/go/src/reflect/type.go:942 +0x135
github.com/containous/yaegi/interp.(*itype).lookupBinField(0xc0003a69c0, 0xc0002c61db, 0x5, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
	/home/fiws/go/src/github.com/containous/yaegi/interp/type.go:602 +0xa9
github.com/containous/yaegi/interp.(*Interpreter).cfg.func2(0xc0002e7900)
	/home/fiws/go/src/github.com/containous/yaegi/interp/cfg.go:1160 +0x399e
github.com/containous/yaegi/interp.(*node).Walk(0xc0002e7900, 0xc000192f70, 0xc000192fa8)
	/home/fiws/go/src/github.com/containous/yaegi/interp/interp.go:113 +0xaf
github.com/containous/yaegi/interp.(*node).Walk(0xc0002e7800, 0xc000192f70, 0xc000192fa8)
	/home/fiws/go/src/github.com/containous/yaegi/interp/interp.go:110 +0x70
github.com/containous/yaegi/interp.(*node).Walk(0xc0002e7600, 0xc000192f70, 0xc000192fa8)
	/home/fiws/go/src/github.com/containous/yaegi/interp/interp.go:110 +0x70
github.com/containous/yaegi/interp.(*node).Walk(0xc0002e7500, 0xc000192f70, 0xc000192fa8)
	/home/fiws/go/src/github.com/containous/yaegi/interp/interp.go:110 +0x70
github.com/containous/yaegi/interp.(*node).Walk(0xc0002e6200, 0xc000192f70, 0xc000192fa8)
[...]

PR #350, opened 2019-09-06
fix:  reuse rather than re-import an already imported source package

mvertes:
For example, when importing "github.com/spf13/cobra", the
package "github.com/spf13/pflag" was imported then parsed
7 times, with all AST resources duplicated.

This change ensures that the source package is imported
only once, and that further import statements will refer to
the already loaded AST.

This change is necessary (but no sufficient) to address #257


mvertes:
Source packages are indexed by package name in interpreter, for symbol resolution. It needs to be indexed by package path instead of package name, as for binary imports. I will fix that in that PR.

PR #382, opened 2019-09-24
interp: don't panic for undefined types

kortschak:
This prevents the runtime panic that otherwise occurs when a symbol that is referenced in a source is not present in the symbol map.

Please take a look.

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -203,7 +203,10 @@ func (interp *Interpreter) cfg(root *node) ([]*node, error) {
 				} else if n.anc.typ != nil {
 					n.typ = n.anc.typ.val
 				}
-				// FIXME n.typ can be nil.
+				if n.typ == nil {
+					err = n.cfgErrorf("undefined type")
+					return false
+				}
 				n.typ.untyped = true
 			}
 			// Propagate type to children, to handle implicit types
--- a/interp/interp_eval_test.go
+++ b/interp/interp_eval_test.go
@@ -368,6 +368,32 @@ func TestEvalChan(t *testing.T) {
 	})
 }
 
+func TestEvalMissingSymbol(t *testing.T) {
+	defer func() {
+		r := recover()
+		if r != nil {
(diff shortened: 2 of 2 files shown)


kortschak:
I was unable to exercise this in the interactive interpreter. However, this will show the effect.

https://play.golang.org/p/teHHMR1Ta4h

```
package main

import (
	"fmt"
	"log"
	"reflect"

)

func main() {
	type Sub struct {
		S string
	}
	type Config struct {
		N int
		F float64
		X Sub
	}
	i.Use(interp.Exports{"xcfg": map[string]reflect.Value{
		"Value": reflect.Zero(reflect.TypeOf(&Config{})),
		// "Sub":   reflect.Zero(reflect.TypeOf(&Sub{})), // Bang!
	}})
	_, err := i.Eval(`import "xcfg"`)
	if err != nil {
		log.Fatal(err)
	}
	v, err := i.Eval(`xcfg.Value{
		N: 5, F: 0.1,
		X: xcfg.Sub{S: "set"},
}`)
	if err != nil {
		log.Fatal(err)
	}

	fmt.Printf("%#v", v)
}
```

If you uncomment the line marked "Bang!" you'll see this otherwise works. Adding the change proposed here results in a returned error rather than a runtime nil-deref panic.

ldez:
Could add the test file: https://github.com/containous/yaegi/tree/master/_test

mvertes:
I need to look at it more closely: you are defining wrappers using `reflect.Zero(reflect.TypeOf(SomeType))` instead of `reflect.ValueOf((*SomeType)(nil))` as normally generated by `goexports`. I need to assess the impact of that.

kortschak:
You can replace the approach I am using with what you have. They are identical in effect.

kortschak:
> Could you add the test file: https://github.com/containous/yaegi/tree/master/_test

This is not appropriate. The code at the reproducer is not for the interpreter, it is for host compiled Go code. I can add a test though.

Issue #317, opened 2019-07-30
typing () gets you stuck in the reply


fiws:
This is a pretty silly bug. But it's still kind of annoying.

Pressing ctrl+c will stop the whole repl session.
This also happens for other chars. For example typing `>` also gets you stuck.

Type this in the repl:
```go
()
```

Expected result:
```console
 What are you even doing?
```

Got:

Me hammering random stuff:
```console
()




Aaargn
()
stahp
quit
EOF
how do i quit vim?
```

mvertes:
Parse error handling in the REPL is rather crude at this moment, resulting in hang, and definitely needs to be improved. A difficulty here is that we have to tolerate incomplete blocks, waiting for a closing '"' or ')' or '}', etc, in the next line, and not fail too early. As a result, we do not fail at all. Hopefully, it is still possible to make better decision based on the precise AST error returned by the AST parser.

fiws:
How about: ctrl+c just clears the input and ctrl+c twice exits? The node.js repl also works like this.

mvertes:
Yes, sounds good, thank you for the suggestion.

kortschak:
When/if #387 is merged, adding signal handling for termination of script execution (and other parts of the interpreter, as described here) will be trivial.

mpl:
This seems to be fixed:

```
% yaegi
> ()
_.go:1:29: expected operand, found ')'
> 
```

So I think we can close.

PR #387, opened 2019-09-25
interp: add EvalWithContext for interpreter cancelation

### LINKED ISSUE #386: interp: add cancellation context to Eval call
#### Proposal

Provide a `context.Context` to the interpreter that can be used for timeout or cancellation.

#### Background

In the case where a code chunk is being executed within an interpreter within a larger application it may be the case that the interpreted code takes too long or will not terminate. In these situations it would be valuable to be able to halt the execution of the interpreter.

Note that it is still possible with a cancelled context for an interpreted chunk to hang if the hang is happening in non-interpreted code called by the interpreted chunk. I am not proposing a solution to this.

#### Workarounds

There is no work around except for running the `yaegi` command as a separate process and using interprocess communication.

kortschak:
Please take a look.

Fixes #386.
Updates #317.


kortschak:
I am lost with this failure. The code failing in `runCfg` simplifies to this when there is no context cancellation.
```
	next := make(chan bltn)
	for exec := n.exec; exec != nil; {
		go func() {
			next <- exec(f)
		}()
		exec = <-next
	}
```
This is essentially equivalent to what was there previously.
```
	for exec := n.exec; exec != nil; {
		exec = exec(f)
	}
```
The latter works (but can't handle the mutex case) and the former gives the failure we see in the travis logs.

```
$ yaegi _test/recover1.go 
hello
panic: test panic

goroutine 11 [running]:
github.com/containous/yaegi/interp._panic.func1(0xc0000d94f0, 0xc00006afb0)
	/home/user/src/github.com/containous/yaegi/interp/run.go:417 +0x7f
github.com/containous/yaegi/interp.runCfg.func2(0xc0000402a0, 0xc0000106c8, 0xc0000d94f0)
	/home/user/src/github.com/containous/yaegi/interp/run.go:115 +0x33
created by github.com/containous/yaegi/interp.runCfg
	/home/user/src/github.com/containous/yaegi/interp/run.go:114 +0x10e

real	0m0.013s
user	0m0.013s
sys	0m0.000s
```

If you can see what it is that I'm missing here, I can fix that. Otherwise this might need to be a TODO and only handle the channel operations and not the generalised blocking case for now.

mvertes:
Also another problem. You are inserting some complex and costly processing at a critical place, in runCfg().

On master:

```console
$ time ./yaegi _test/l3.go
[...]
19857408

real    0m1.528s
user    0m1.514s
sys     0m0.017s
```

On kortschak-eval/context:

```console
$ time ./yaegi _test/l3.go
[...]
19857408

real    2m26.237s
user    2m38.157s
sys     0m16.658s
```

I wonder why you didn't keep your first attempt, more interesting to me, which was not spawning a goroutine a each and every instruction to execute

kortschak:
Yes, that's why I originally split it into progressive changes so I could back them out. I was concerned about the cost, but without that there is no way to be able to cancel from non-language level blocking operation like `mu.Lock()` or even just long calls to compiled code.

However, at this point, I think it's probably more important to have it at least partially working and then fix those issues later. If you agree, I'll revert the last commit here and leave it as only handling channel operations.

There is a non-goroutine approach that I had hoped would work, but it does not allow the blocking calls to be cancelled.
```
	done := make(chan struct{})
	close(done)
	for exec := n.exec; exec != nil; {
		select {
		case <-ctx.Done():
			return
		case <-func() chan struct{} {
			exec = exec(f)
			return done
		}():
		}
	}
```

If I remove the `go` keyword from the minimal reproducer, it stops failing. So, there is clearly a concurrency issue here. What seems to be happening is that even though `_recover` is called, the `bltn` that is built with that call is never executed.

mvertes:
I haven't spent time yet on making the interpreter interruptible, so thank you for tackling this worthwhile problem. The approach you took in the first version of your patch, even if partial, seems interesting and promising. To me the first priority would be to break the execution loop and release the CPUs. Other situations where the interpreter is blocked only result in a limited resource leak, not as worse as a spinning program, so it's already a significant progress.

kortschak:
No worries. I'll revert the last commit.

Note that a double lock will not return from a cancellation with that state of the code; it's not just a resource leak, but a complete stall.

kortschak:
w00t!

interp/run.go
```diff
@@ -852,20 +864,37 @@ func callBin(n *node) {
...
+				}
 				copy(f.data[n.findex:], out)
 				return tnext
 			}
 		}
 	}
 }
 
-func getIndexBinMethod(n *node) {
+func callWithContext(ctx context.Context, fn reflect.Value, in []reflect.Value) (out []reflect.Value, ok bool) {
```
mvertes: I think that `callWithContext` suffers from the same performance problem that was exposed in your previous change of `runCfg`: you add a goroutine+select for each call of a binary function!!! In addition, it's a kind of useless as we have no way to interrupt the binary call itself. If it is a matter of not stalling the interpreter as a whole, we could just spawn just one goroutine in the top `eval` call instead.

mvertes:
I would like to proceed step by step. We still have to assess the performance in some of our internal (at containous)  benchmarks involving real apps. The best candidate for merge as for now would be your version without the last commit.

interp/run.go
kortschak: >If it is a matter of not stalling the interpreter as a whole, we could just spawn just one goroutine in the top eval call instead.

I think this is a good solution. If we have the per exec check that's in the first commit for live-lock and a global goroutine/select for dead lock then the code changes are less than here and there is only one goroutine/select. I doubt that the performance impact it likely to be as bad a your are worried about, but this is a better solution.

kortschak:
PTAL

I have the global goroutine/select in, which should give O(0) performance impact beyond the per-instruction cost of the first commit. Also much less and clearer code.

kortschak:
This failure is not related.

kortschak:
Maintaining this as a non-conflicting patch against master is becoming onerous; pretty much every commit to cfg or run will have either a lexical or a semantic conflict with this set of changes.

mvertes:
Yes, but I plan to switch to v0.7 very soon. Just released v0.6.2, and now ready to move forward

kortschak:
OK. Thanks for the update.

mvertes:
With the last change , we have cancelled the overhead in most of our benchmarks

kortschak:
Please let me think about this for a while before you merge.

kortschak:
I think this number should probably be user-configurable. In my use cases, there are expensive binary calls that would be run as a sequence in interpreted code in an interactive interpreter. Not returning for 1000 of these after a cancellation would not be acceptable.

The question is where to put the configuration. The two options are in the `interp.Option` or alternatively in the `context.Context` using thw `Value` accessor.

mvertes:
You are right. 1000 is probably overkill, the point here was to verify the possibility to cancel the overhead for the worst case introduced by this change, where a single instruction is relatively fast compared to the cost of checking the cancel condition. When a single instruction (a binary call) takes much longer, then of course repeating 1000 times the instruction defeats the purpose.

Scratching my head to see if we can come with something simpler, safe and more efficient.

kortschak:
It can be merged this way in the interim. I'm happy to do some exploration of approaches. I think I'm leaning towards the `context.Context` approach since that keeps the related concerns together.

mvertes:
I think the current patch is wrong. There is a subtle design flaw in the way contexts are used. The cancellation `ctx` should flow during the run phase only, not during the closure generation. 

In the current situation, as `ctx` is passed during closure generation, and closures are run potentially multiple times, we will end up cancelling background goroutines started in the past, in addition to the ones that we want to stop, because they will share the same `ctx` inherited from closure caller context.

The generated closures have a current signature of `type bltn func(*frame) bltn`, see [here]( https://github.com/containous/yaegi/blob/1cf327bd7d1d1ec88cb239fd5bcedd8f645c10c0/interp/run.go#L12). This type should become `type bltn func(*context.Context, *frame) bltn`, and so on. In `runCfg()`, we should have something like `exec = exec(ctx, f)` instead of `exec = exec(f)`.

The good news is that the correct patch should be less intrusive on master than this one, as only the run exec generators have be changed, without the need to propagate `ctx` through GTA, CFG, scopes and types.

kortschak:
In order to get `exec = exec(ctx, f)`, we need to have `runCfg(ctx context.Context, n *node, f *frame)` The threading of the context through the call chain was a light as I could make it and have the package compile. The existence of the context during the CFG, type evaluation and so on is a purely mechanical consequence of this. Adding a context parameter to the `bltn` signature and checking in those will make a bigger set of changes, in addition to the change that exist here and will need to continue to exist because of the call chain.

Note that there is only one location in the running interpreter than check for the context having cancelled, and one outside the interpreter that brings things down for non-cancelable running code. There is nothing in the closures that responds to this at this stage.

mvertes:
I look at the places where `runCfg` is called from. If you take `genFunctionWrapper` for example, with the change of  `bltn` type, the ctx does not require anymore to be taken from an argument of `genFunctionWrapper`, thus removing the need to propagate through CFG. The same arises for `call` and `callBin`. As a result the context can be flowed directly from `evalWithContext` to `run`, to `runCfg`.

kortschak:
If you change `bltn` to `func(f *frame) bltn` then the call to `n.exec` [here](https://github.com/containous/yaegi/blob/1cf327bd7d1d1ec88cb239fd5bcedd8f645c10c0/interp/run.go#L112) needs access to the context.

In order to get a `context.Context` argument to `exec`, it must come from the signature of `runCfg`. For a call to `runCfg` to have that, it must have access to is via the `*interp.run` signature *and* some aspect of the `genFunctionWrapper` signature. This is either going to come directly as an argument of `genFunctionWrapper` or as a field of `*node`. The latter is not recommended practice for passing of contexts. The rest flows naturally.

The alternative is to close the context into the `bltn` closures (not recommended since contexts should be passed explicitly), then every function that assigns to `n.exec` must have access to the context and we are again back to where we are with this PR, just with a more opaque passing strategy.

I invite you to work through the proposal you have. It's not nearly as simple as you are suggesting.

In addition to this, the cancellation with the context being threaded though the preparatory stages is actually a benefit; the code in this PR does not hook cancellations in these phases, but could, this is why I referenced #317.

mvertes:
I explored a bit more forwarding the context into the n.exec closures, it's not so simple as you said, and I ended up with more overhead, having to pass context to all value getter closures as well.

I decided to experiment another way, which is to not use the context package and instead to leverage on the existing interpreter execution infrastructure, the frames (which in the end acts as passed contexts), to manage the cancellation. See PR #401 for further details.

mvertes:
Superseded by #402

Issue #315, opened 2019-07-30
repl: improved UI


sbinet:
it would be great (beside having regular "readline-like" capabilities like alluded to in #286) to have support for:

- tab-completion
- up-/down-arrow history search

in [neugram](https://neugram.io), we used [github.com/peterh/liner](https://github.com/peterh/liner).

midnightrun:
Tab Completion would be awesome.

gmarik:
got working prototype in https://github.com/containous/yaegi/pull/389

blasrodri:
I added a very simple patch for taking the previous command when pressing the right arrow.
It's far from being ready for a PR. But I just wanted to know if this makes any sense at all.

```diff
diff --git a/interp/interp.go b/interp/interp.go
index 6111a73..2d3361f 100644
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -2,6 +2,7 @@ package interp
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"fmt"
 	"go/build"
@@ -138,6 +139,7 @@ type Interpreter struct {
 	srcPkg   imports           // source packages used in interpreter, indexed by path
 	pkgNames map[string]string // package names, indexed by path
 	done     chan struct{}     // for cancellation of channel operations
+	commands []string
 }
 
 const (
@@ -496,7 +498,19 @@ func (interp *Interpreter) REPL(in io.Reader, out io.Writer) {
 	src := ""
 	s := bufio.NewScanner(in)
 	for s.Scan() {
-		src += s.Text() + "\n"
+		n := len(interp.commands)
+		if bytes.Equal(s.Bytes(), []byte{27, 91, 65}) {
+			if n > 0 {
+				src = interp.commands[n - 1]
+				// shrink it
+				interp.commands = interp.commands[:n - 1]
+			} else {
+				src = ""
+			}
+		} else {
+			src += s.Text() + "\n"
+			interp.commands = append(interp.commands, src)
+		}
 		ctx, cancel := context.WithCancel(context.Background())
 		defer cancel()
 		handleSignal(ctx, cancel)
```

mpl:
Hi @blasrodri 

We think that if we're ever going to add some readline type capabilities, we'll probably go with an existing full-blown library, such as liner, because:
1) If we go with a small subset of features, there's always going to be someone who is going to want that we support their favorite one. It's easier to say "you get everything that is in liner, if you want anything else, ask liner to implement it".
[...]

PR #389, opened 2019-09-25
spike: yaegi readline

gmarik:
A PR to gauge interest of adding a  `readline` or alike  functionality to the cli-command.
Note: code is a throwaway prototype to see how difficult it is to integrate [liner](https://github.com/peterh/liner): a compact `readline` implementation.

## problem
`yeagi` without common `readline`  functionality is a poor UX and may limit wider adoption.

## how: in action
1. `cd cmd/yaegi-readline/`
2. `go1.11 run ./`
3. enjoy: history, shortcuts, completion, etc…

## PS
met @emilevauge at the docker meetup  yesterday and he suggested going ahead and opening  a PR. Previously I thought you're not interested in adding any dependencies.


ldez:
Hello,

Thanks, we don't want to introduce external dependencies

> Written in pure Go, using only the standard library

https://github.com/containous/yaegi#features

gmarik:
> we don't want to introduce external dependencies.

IMO it's a case when having a dependency is justified, since:
0. efforts to re-implement from scratch are significant
0. it's a separate, terminal/leaf package: cli dependencies will not "pollute" the core
0. end user may choose whether to use `yaegi` or `yaegi-readline`

I'd love to do it but i'll take some time unless short-cuts are taken…

emilevauge:
Hey @gmarik :)
Thanks for the PR, you were fast!
In fact, I think I suggested to create an issue before, in order to open the discussion, as I was not aware of any potential reason to not implement this.
@ldez I understand your statement, but can we try to find a consensus without rewriting everything (which, I hope you agree, does not make sense)?
IMHO, either we provide a rich REPL, either we don't provide anything.

ldez:
In the future, I think we will split the yaegi core and the yaegi CLI/REPL as 2 projects.

When we will do that we will be able to use external dependencies.

For now, there are some side effects to add an external dependency.

The issue related #315.

gmarik:
1. hey emile sorry i think i didnt catch the bit about the “issue”. either way my thinking was  a working prototype is better than “an idea” since there’s already one
2. imo splitting is unnecessary: library users will not depend on readline as long as it’s separated and isolated

let me know your thoughts

mvertes:
As explained previously, for the moment we would like to avoid external dependencies, as we don't have any yet. We believe that a good advanced REPL could be built on top of yaegi in an external project.

Thanks.

PR #392, opened 2019-09-28
interp: fix range expression handling

### LINKED ISSUE #340: for loop edge case freezes yaegi

```go
package main

import "fmt"

func main() {
        v := []int{1, 2, 3}
        for i := range v {
                v = append(v, i)
        }
        fmt.Println(v)
}
}
```

Expected result:
```console
$ go run ./sample.go
// ouput
[1 2 3 0 1 2]
```

Got:
```console
$ yaegi ./sample.go
// ouput
```
There is no output, it seems to get stuck

kortschak:
Please take a look.

Updates #340.

### DIFF
--- a/interp/run.go
+++ b/interp/run.go
@@ -1534,11 +1534,11 @@ var rat = reflect.ValueOf((*[]rune)(nil)).Type().Elem() // runes array type
 
 func _range(n *node) {
 	index0 := n.child[0].findex // array index location in frame
+	index2 := index0 - 1        // shallow array for range, always just behind index0
 	fnext := getExec(n.fnext)
 	tnext := getExec(n.tnext)
 
 	var value func(*frame) reflect.Value
-	var a reflect.Value
 	if len(n.child) == 4 {
 		an := n.child[2]
 		index1 := n.child[1].findex // array value location in frame
@@ -1548,6 +1548,7 @@ func _range(n *node) {
 			value = genValueRangeArray(an)
 		}
 		n.exec = func(f *frame) bltn {
+			a := f.data[index2]
 			v0 := f.data[index0]
 			v0.SetInt(v0.Int() + 1)
 			i := int(v0.Int())
@@ -1567,7 +1568,7 @@ func _range(n *node) {
 		n.exec = func(f *frame) bltn {
 			v0 := f.data[index0]
 			v0.SetInt(v0.Int() + 1)
-			if int(v0.Int()) >= a.Len() {
+			if int(v0.Int()) >= f.data[index2].Len() {
 				return fnext
 			}
 			return tnext
@@ -1577,8 +1578,8 @@ func _range(n *node) {
 	// Init sequence
 	next := n.exec
 	n.child[0].exec = func(f *frame) bltn {
-		a = value(f)
-		f.data[index0].SetInt(-1)
+		f.data[index2] = value(f) // set array shallow copy for range
+		f.data[index0].SetInt(-1) // assing index value
 		return next
 	}
 }
--- a/interp/run.go
+++ b/interp/run.go
@@ -1538,16 +1538,16 @@ func _range(n *node) {
 	tnext := getExec(n.tnext)
 
 	var value func(*frame) reflect.Value
+	var a reflect.Value
 	if len(n.child) == 4 {
 		an := n.child[2]
 		index1 := n.child[1].findex // array value location in frame
 		if isString(an.typ.TypeOf()) {
 			value = genValueAs(an, rat) // range on string iterates over runes
 		} else {
-			value = genValueArray(an)
+			value = genValueRangeArray(an)
 		}
 		n.exec = func(f *frame) bltn {
-			a := value(f)
 			v0 := f.data[index0]
 			v0.SetInt(v0.Int() + 1)
 			i := int(v0.Int())
@@ -1562,10 +1562,9 @@ func _range(n *node) {
 		if isString(an.typ.TypeOf()) {
 			value = genValueAs(an, rat) // range on string iterates over runes
 		} else {
-			value = genValueArray(an)
+			value = genValueRangeArray(an)
 		}
 		n.exec = func(f *frame) bltn {
-			a := value(f)
 			v0 := f.data[index0]
 			v0.SetInt(v0.Int() + 1)
 			if int(v0.Int()) >= a.Len() {
@@ -1578,6 +1577,7 @@ func _range(n *node) {
 	// Init sequence
 	next := n.exec
 	n.child[0].exec = func(f *frame) bltn {
+		a = value(f)
 		f.data[index0].SetInt(-1)
 		return next
 	}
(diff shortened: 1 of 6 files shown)


interp/run.go
```diff
@@ -1538,16 +1538,16 @@ func _range(n *node) {
 	tnext := getExec(n.tnext)
 
 	var value func(*frame) reflect.Value
+	var a reflect.Value
```
mvertes: No. You are using this variable to store live data computed during interpretation (the actual array). It has to be stored in the frame. The frame position and type has to be allocated during CFG.

The flaw in my implementation of range is that only 1 register is used to store the range index (store at index0 location in frame). We should use a second register to store the range limit (an integer), computed once during the init sequence. Then the test before index increment should be done against this limit value instead of array length (which can change). No need for genValueRangeArray. But need to allocate a second register for storing range limit ans remember its location in say index1. The allocation has to be done with a `sc.Add(limitType)`  (where limit type is an `int`) around  https://github.com/containous/yaegi/blob/2c2b471cb91f520b10f9ea8e0f988de221c94b6a/interp/cfg.go#L106

interp/run.go
```diff
@@ -1538,16 +1538,16 @@ func _range(n *node) {
...
 	if len(n.child) == 4 {
 		an := n.child[2]
 		index1 := n.child[1].findex // array value location in frame
 		if isString(an.typ.TypeOf()) {
 			value = genValueAs(an, rat) // range on string iterates over runes
 		} else {
-			value = genValueArray(an)
```
mvertes: no change here, see above explanation

interp/run.go
```diff
@@ -1562,10 +1562,9 @@ func _range(n *node) {
...
```
mvertes: no change here as well, see above

interp/run.go
kortschak: I don't think so. A range on an array results in a copy of the array. If you don't do this you will not execute the same way as the Go spec and you will fail the range1.go test.

interp/run.go
```diff
@@ -1578,6 +1577,7 @@ func _range(n *node) {
 	// Init sequence
 	next := n.exec
 	n.child[0].exec = func(f *frame) bltn {
+		a = value(f)
```
mvertes: we should have a frame location (say index1) to store the array length:

```go
v1 := f.data[index1]
v1.SetInt(value(f).Len())
```

interp/value.go
```diff
@@ -118,6 +118,22 @@ func genValueArray(n *node) func(*frame) reflect.Value {
 	return value
 }
 
+func genValueRangeArray(n *node) func(*frame) reflect.Value {
```
mvertes: no need for this function

kortschak: See previous comment.

interp/run.go
mvertes: Right, `range1.go` demonstrates that, although I could not find a clear mention in the spec. I agree with your implementation using `genValueArray` except its value has to be stored in the frame.

interp/run.go
kortschak: As pointed out, this will still fail to match the compiled code's behaviour. Making a new value that is outside the scope of the interpreted code is far more aligned with the compiled code case where the range expression is not available to the Go program using anything except unsafe access.

In both the slice and array cases the value of the range expression is copied (it is dangerous to think of arrays and slices as arrays since this leads to misconceptions of shared behaviours). In the case of a slice, we have approximately what you are suggesting while in the array case you get an entire copy of the array. Now you could differentiate and copy for the array and keep the length for the slice, but we already get that behaviour with the change here.

interp/run.go
kortschak: I don't understand why it needs to be in the frame. In the gc, there is no programmatic access to the value of the range expression after it has been evaluated. As far as the user code is concerned the value is a magic thing that exists in the æther. That is replicated here.

mvertes: If the loop is running in goroutines, the same memory location to store the array will be used by concurrent tasks (as the _range generator is run only once per code chunk), as opposed to a local stack in compiled go. Using the frame ensures the same behaviour as in compiled go.

mvertes: As the frame is the works a local stack within the interpreter

kortschak: OK. This makes sense. I'll give it a go. How do I make a slot in the frame, it's not immediately obvious to me how to do this.

This is what I have tried. It fails.

```diff
diff --git a/interp/cfg.go b/interp/cfg.go
index be1e43e..8b6665d 100644
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -98,7 +98,13 @@ func (interp *Interpreter) cfg(root *node) ([]*node, error) {
                                        case stringT:
                                                ktyp = sc.getType("int")
                                                vtyp = sc.getType("rune")
-                                       case arrayT, variadicT:
+                                       case arrayT:
+                                               otyp := o.typ
+                                               oindex := sc.add(otyp)
+                                               sc.sym[o.ident] = &symbol{index: oindex, kind: varSym, typ: otyp}
+                                               o.findex = oindex
+                                               fallthrough
+                                       case variadicT:
                                                ktyp = sc.getType("int")
                                                vtyp = o.typ.val
                                        }
diff --git a/interp/run.go b/interp/run.go
index f04ae9a..b7e7295 100644
--- a/interp/run.go
+++ b/interp/run.go
@@ -1534,11 +1534,11 @@ var rat = reflect.ValueOf((*[]rune)(nil)).Type().Elem() // runes array type
 
 func _range(n *node) {
        index0 := n.child[0].findex // array index location in frame
+       index1 := n.child[1].findex // range expression
        fnext := getExec(n.fnext)
        tnext := getExec(n.tnext)
 
        var value func(*frame) reflect.Value
-       var a reflect.Value
        if len(n.child) == 4 {
                an := n.child[2]
                index1 := n.child[1].findex // array value location in frame
@@ -1549,6 +1549,7 @@ func _range(n *node) {
                }
                n.exec = func(f *frame) bltn {
                        v0 := f.data[index0]
+                       a := f.data[index1]
                        v0.SetInt(v0.Int() + 1)
                        i := int(v0.Int())
                        if i >= a.Len() {
@@ -1566,6 +1567,7 @@ func _range(n *node) {
                }
                n.exec = func(f *frame) bltn {
                        v0 := f.data[index0]
+                       a := f.data[index1]
                        v0.SetInt(v0.Int() + 1)
                        if int(v0.Int()) >= a.Len() {
                                return fnext
@@ -1577,8 +1579,8 @@ func _range(n *node) {
        // Init sequence
        next := n.exec
        n.child[0].exec = func(f *frame) bltn {
-               a = value(f)
                f.data[index0].SetInt(-1)
+               f.data[index1].Set(value(f))
                return next
        }
 }
```

interp/run.go
mvertes: Agree. No need to store just the index as we have to get access to the entire array copy anyway

mvertes:
We can now merge this PR. I found other checks to do in CFG about range statement, but they can be added later, and should not affect the logic in _range.

kortschak:
Thanks. I can use your addition to figure out how to add the similar requirements for in-frame storage of of the iterator state for the map range fix PR #393.

PR #393, opened 2019-09-28
interp: fix map range handling

kortschak:
Please take a look.

Note that this change requires dropping Go1.11 support.

Updates #340.

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -87,6 +87,8 @@ func (interp *Interpreter) cfg(root *node) ([]*node, error) {
 						}
 					case mapT:
 						n.anc.gen = rangeMap
+						ityp := &itype{cat: ptrT, rtype: reflect.TypeOf((*reflect.MapIter)(nil))}
+						sc.add(ityp)
 						ktyp = o.typ.key
 						vtyp = o.typ.val
 					case ptrT:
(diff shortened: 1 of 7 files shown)


kortschak:
The second commit here puts the `reflect.MapIter` into the frame. I'm not sure if this is how you would like it done. I could make a pre-defined symbol in `initUniverse`, but I thought I'd wait for your input.

kortschak:
This failure is not related.

kortschak:
Can you explain how the `cat` field works?

mvertes:
To be merged in v0.7.0 because of go1.11 dropping

interp/cfg.go
```diff
@@ -87,7 +87,7 @@ func (interp *Interpreter) cfg(root *node) ([]*node, error) {
 						}
 					case mapT:
 						n.anc.gen = rangeMap
-						ityp := &itype{cat: ptrT, rtype: reflect.TypeOf((*reflect.MapIter)(nil))}
+						ityp := &itype{cat: valueT, rtype: reflect.TypeOf((*reflect.MapIter)(nil))}
```
kortschak: Can you explain this change?

mvertes: The type here refers to a binary type (vs interpreter defined), from `reflect` package, the right category is valueT, For ptrT category, the data is stored in `val` vs `rtype`. It's for consistency, as `rtype` being filled, the type is usable anyway.

kortschak: OK. Thanks.

PR #399, opened 2019-10-01
feature: rename exported func Repl into REPL

### LINKED ISSUE #390: interp: nit: REPL is spelled wrong
This is a tiny thing, but since the code is still apparently in pre-alpha, it's when it could be addressed if ever.

Per [the standard Go style](https://github.com/golang/go/wiki/CodeReviewComments), [acronyms should be entirely uppercase](https://github.com/golang/go/wiki/CodeReviewComments#initialisms). Thus `interp.Repl` should be spelled `interp.REPL`. Note that the all upper case spelling is used in discussion in the issues and PRs by containous team members so apart from code change costs this is likely to be uncontroversial.

mvertes:
Fixes #390

### DIFF
--- a/.golangci.toml
+++ b/.golangci.toml
@@ -44,3 +44,10 @@
   [[issues.exclude-rules]]
     path = "interp/.+_test\\.go"
     linters = ["goconst"]
+
+  [[issues.exclude-rules]]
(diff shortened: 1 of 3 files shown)


mvertes:
LGTM

PR #401, opened 2019-10-04
feature: add Stop() and EvalTimeout()

mvertes:
This is an alternate solution to `context.Context`, to implement
eval with timeout and cancellation, as per #386

This solution uses a `runid` counter added both to the execution
frames and the global `Interpreter` struct. When Eval starts, the
current `runid` from `Interpreter` is propagated to the frames
(a new frame is created at each function call, including closures
and goroutines; a frame inherits the `runid` from its ancestor).
in frames, the runid is written once at frame init, and never modified.
The interpreter runid can be modified to trigger the stop condition
(see Stop()).

Eval stop is triggered when the current frame runid is different
from the interpreter runid. This stop condition is evaluated in the
execution loop of `runCfg` at each instruction step. From my measures,
this integer equality test is 20 times faster than a select statement
on closed channel.

An interpreter session consists of multiple evals. Eval can spawn
a set of background goroutines. With this algorithm, all active
goroutines will be stopped, no matter which previous eval started
them. After a stop, a new Eval can happen.

To be more complete, a mechanism to unblock channels read/write, select
and locks must be added as well, this is planned in a further change.
This can be similar to a closed channel broadcast (as in context.Context)
except that the stop channel will be probably defined once in Interpreter
struct, and be used only for blocking calls, not execution loop. PR #387
proposed at some point an implementation of such interruptible blocking
calls.

The unit tests have been copied from PR #387.

the REPL has been modified to allow Eval interrupt by CTRL-C using
the above mechanism.

This is experimental at this point, and for discussion and comparison
with #387.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -324,6 +324,8 @@ func (interp *Interpreter) Eval(src string) (reflect.Value, error) {
 // Stop stops the execution of the interpreter
 func (interp *Interpreter) Stop() { interp.runid++ }
 
+// EvalTimeout is the same as Eval, except it will return after a timeout it the
+// execution is not complete.
 func (interp *Interpreter) EvalTimeout(src string, d time.Duration) (v reflect.Value, err error) {
 	done := make(chan struct{})
 
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -7,8 +7,10 @@ import (
 	"go/scanner"
 	"go/token"
 	"os"
+	"os/signal"
 	"reflect"
 	"strconv"
+	"time"
 )
 
 // Interpreter node structure for AST and CFG
@@ -46,12 +48,15 @@ type receiver struct {
 	index []int         // path in receiver value for interface or value type
 }
 
+type runid uint
+
 // frame contains values for the current execution level (a function context)
 type frame struct {
 	anc       *frame            // ancestor frame (global space)
 	data      []reflect.Value   // values
 	deferred  [][]reflect.Value // defer stack
 	recovered interface{}       // to handle panic recover
+	runid                       // for cancellation
 }
 
 // Exports stores the map of binary packages per package path
@@ -80,6 +85,7 @@ type Interpreter struct {
 	binPkg   Exports           // binary packages used in interpreter, indexed by path
 	srcPkg   imports           // source packages used in interpreter, indexed by path
 	rdir     map[string]bool   // for src import cycle detection
+	runid                      // for cancellation
 }
 
 const (
@@ -296,6 +302,7 @@ func (interp *Interpreter) Eval(src string) (reflect.Value, error) {
 		return res, err
 	}
 	interp.resizeFrame()
+	interp.frame.runid = interp.runid
 	interp.run(root, nil)
 
 	for _, n := range initNodes {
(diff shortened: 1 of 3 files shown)


kortschak:
This is racey. Also note that a `context.Context` could be used with this underlying implementation to give a more generically usable system, eliding the need for the `Stop` method.

For reference, the chan operation cancellation changes that work with #387 is [here](https://github.com/kortschak/yaegi/commit/2fc203e4255c6e31c0f11bd33de9b04cde169050). That approach will not work here without adding essentially all the other background in #387, unless as proposed above a stop chan is added to the interpreter. It's unclear to me after a brief read how safe that will be.

interp/interp.go
```diff
@@ -314,6 +321,28 @@ func (interp *Interpreter) Eval(src string) (reflect.Value, error) {
...
+func (interp *Interpreter) EvalTimeout(src string, d time.Duration) (v reflect.Value, err error) {
+	done := make(chan struct{})
+
+	go func() {
+		defer close(done)
+		v, err = interp.Eval(src)
+	}()
+
+	select {
+	case <-time.After(d):
```
kortschak: Make this a `ctx.Done()` call, then cancellation, timeouts and the safer time cancellation approach that is deadlines.

interp/interp.go
```diff
@@ -314,6 +321,28 @@ func (interp *Interpreter) Eval(src string) (reflect.Value, error) {
 	return res, err
 }
 
+// Stop stops the execution of the interpreter
+func (interp *Interpreter) Stop() { interp.runid++ }
```
kortschak: This needs to update an atomic or be locked somehow. In general, people will want to run things in goroutines, but even here as this is implemented there is a race between the write to `interp.runid` here and the read in the `run` method.

I would suggest having this method unexported and have the user interact with it via a context as stated elsewhere.

interp/interp.go
```diff
@@ -80,6 +85,7 @@ type Interpreter struct {
 	binPkg   Exports           // binary packages used in interpreter, indexed by path
 	srcPkg   imports           // source packages used in interpreter, indexed by path
 	rdir     map[string]bool   // for src import cycle detection
+	runid                      // for cancellation
```
kortschak: This needs to be protected from races. This will add synchronisation overhead, but will be less than for the select approach.

interp/interp.go
```diff
@@ -337,6 +366,14 @@ func (interp *Interpreter) Repl(in, out *os.File) {
 	prompt := getPrompt(in, out)
 	prompt()
 	src := ""
+	c := make(chan os.Signal)
+	signal.Notify(c, os.Interrupt)
+
+	go func() {
```
kortschak: This approach is one-shot. See https://github.com/kortschak/yaegi/commit/431f838ad4574f74d9ddba1245ab3de6b83e338e for an alternative.

interp/interp_eval_test.go
```diff
@@ -396,6 +396,83 @@ func TestEvalMissingSymbol(t *testing.T) {
...
+			i := interp.New(interp.Options{})
+			i.Use(stdlib.Symbols)
+			_, err := i.Eval(`import "sync"`)
+			if err != nil {
+				t.Errorf(`failed to import "sync": %v`, err)
+				return
+			}
+			_, err = i.EvalTimeout(src, 100*time.Millisecond)
+			if err == nil {
+				t.Errorf("should have failed")
```
kortschak: This should not be `Errorf`.

kortschak:
Alternative, with the races fixed and chan operation cancelation added is at #402. The stop method is unexported since it now explodes if called twice. Given that `context.Context` is a more powerful hook and is common if network applications I don't see a problem with this.

mvertes:
Closing it, as work has been pursued in  #402

PR #402, opened 2019-10-05
interp: add eval cancelation by semaphore

kortschak:
Please take a look.


kortschak:
Note that this is still racey. The glaringly obvious races are gone, but that leaves now only the subtle and difficult find races. This kind of problem is why channels were included in Go.

To demonstrate the race, run the interpreter in interactive mode after having built it with `-race`. Execute `for {}` and then <kbd>Ctrl</kbd>-<kbd>C</kbd> to cancel. Everything should be OK, but then press <kbd>Enter</kbd>. You will see a large number of data race complaints. These are all buried in the corpse of the previous evaluation. Ensuring all of these are made concurrency-safe looks hairy.

mvertes:
Thanks for your hard work on this!

The race detector does not complain anymore for the simple REPL test that you mentioned, All the fixed races were happening during init/compiling phase, not execution, so negligible impact on performances. I havent tested much deeper yet. I assume that some substantial work remains on that front.

I will shortly commit some benchmarks as well. While the execution loop is not too much impacted, channel operations are highly slowed down. You can see by running `_test/sieve.go` with and without cancellation. It's not clear to me yet why it is so.

interp/interp.go
```diff
@@ -57,6 +58,7 @@ type frame struct {
 	recovered interface{}       // to handle panic recover
 	runid     uint64            // for cancellation
 	done      chan struct{}     // for cancellation of channel operations
+	mutex     sync.Mutex
```
kortschak: This should be placed above the fields that is protects, with an annotation listing those fields it is protecting.

Also, I'd suggest this be a `RWMutex` since that will save most contended cases.

interp/scope.go
```diff
@@ -164,6 +164,8 @@ func (interp *Interpreter) initScopePkg(n *node) (*scope, string) {
 		pkgName = p.child[0].ident
 	}
 
+	interp.mutex.Lock()
+	defer interp.mutex.Unlock()
```
kortschak: This can be a direct call to unlock since there's only a single exit here, and it's close. This would save on defer cost.

kortschak:
Yes, I'd expect channel operations to be slower. Select is more costly than a pure chan send or receive. So we would expect selects to stay approximately the same, but a quantum shift for the change from chan send/recv since they are now all selects.

The `sieve.go` is cute, but not a good implementation. I would not be optimising for that over correctness.

kortschak:
The last commit can be reverted or put somewhere else, but this seemed like a reasonable place to put it. I have written it as a fork/exec so that it doesn't bring down the entire test build in the case of failure.

interp/interp.go
kortschak: Actually this should have been a comment on `Interpreter`.

kortschak:
I've written a lighter test for this. In place of the full yaegi command. This actually tests things better and not shows the presence of races still. I'm working on this.

interp/run.go
```diff
@@ -2036,10 +2034,16 @@ func recv(n *node) {
 	if n.fnext != nil {
 		fnext := getExec(n.fnext)
 		n.exec = func(f *frame) bltn {
-			chosen, v, _ := reflect.Select([]reflect.SelectCase{
-				{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(f.done)},
-				{Dir: reflect.SelectRecv, Chan: value(f)},
-			})
+			ch := value(f)
+			// Fast: channel read doesn't block
+			if x, ok := ch.TryRecv(); ok {
```
kortschak: I don't think this is the right semantics. It changes completely how channel operations work.

interp/run.go
```diff
@@ -2033,43 +2033,63 @@ func recv(n *node) {
...
-			if x, ok := ch.TryRecv(); ok {
-				if x.Bool() {
+	if n.interp.cancelChan {
+		// Cancellable channel read
+		if n.fnext != nil {
+			fnext := getExec(n.fnext)
+			n.exec = func(f *frame) bltn {
+				ch := value(f)
+				// Fast: channel read doesn't block
+				if x, ok := ch.TryRecv(); ok {
```
kortschak: I don't think this is correct.

interp/run.go
mvertes: What is not correct exactly ? I don't see. If `TryRecv` succeeds, no need to `Select` and block, no ?

kortschak: Sorry, you are absolutely correct. I misread the `TryRecv` docs.

interp/run.go
```diff
@@ -2036,10 +2034,16 @@ func recv(n *node) {
...
 		fnext := getExec(n.fnext)
 		n.exec = func(f *frame) bltn {
-			chosen, v, _ := reflect.Select([]reflect.SelectCase{
-				{Dir: reflect.SelectRecv, Chan: reflect.ValueOf(f.done)},
-				{Dir: reflect.SelectRecv, Chan: value(f)},
-			})
+			ch := value(f)
+			// Fast: channel read doesn't block
+			if x, ok := ch.TryRecv(); ok {
+				if x.Bool() {
```
kortschak: Curious. How do we know `x` represents a `bool`? Shouldn't this be.
```
if x, ok := ch.TryRecv(); ok {
    return tnext
}
```
and falling through in the failing case?

interp/interp.go
```diff
@@ -428,6 +426,7 @@ func (interp *Interpreter) Repl(in, out *os.File) {
 	prompt := getPrompt(in, out)
 	prompt()
 	src := ""
+	interp.opt.cancelChan = true
```
kortschak: Please don't do this. This makes cancellation outside the yeagi REPL method unavailable. This in turn means that other REPL, say using peterh/liner cannot cancel, and particularly important for me (i.e. why I'm working on this), that cancellation of `Eval` calls is not possible.

interp/run.go
mvertes: When`fnext`  is not `nil`, we are in presence of a  conditional branch (to `tnext` of `fnext`), which occurs only in case of boolean value. This has been detected and set during CFG.  We have to test the content of `x` which contains the boolean value. `ok` only carries if the chan read completed, not the data itself.

kortschak: OK

interp/interp.go
mvertes: Ok. An exported method on Interpreter object, or making `CancelChan` an exported field of interpreter would work for you then ?

kortschak: I'm not a fan of masses of knobs; too much time in the Go ecosystem and seeing people abuse them in other contexts.

The issue here is the choice of the default; I think that the fast option should not be the default since it leaves the unaltered interpreter (the value returned by `interp.New` with default options) in a state that leaks goroutines. This will be a constant source of issues for you and for client code that uses the interpretter embedded. Then what to call it? It needs to be a name that indicates both fast and unsafe (which it is).

My preferred option would be an environment variable (probably undocumented), `YAEGI_FAST_CHAN`, that people who want to use this can if they dig hard enough; it should not be encouraged.

mvertes: As a side note, no matter the value `cancelChan`, and the use of cancellable or blocking chan read / write, `EvalWithContext` will always return and work as expected for integration within a custom REPL. You already did everything for that. What changes is that some background goroutines may remain hanged on a chan in one case, or released in another, but no change from `EvalWithContext` point of view. Or am I missing something ?

kortschak: Yeah, I understand that. I'm concerned about the goroutine leaks and resource depletion attack vectors.

kortschak:
Thank you for working with me on this. It has been very helpful and enjoyable.

kortschak:
Will this be added to the v0.7.x milestone in place of #387?

mvertes:
I hope so. Allow a few days, thank you for your patience.

PR #404, opened 2019-10-07
interp: use io.Reader and io.Writer for REPL parameters

### LINKED ISSUE #403: interp: consider making REPL take io.Reader and io.Writer rather than *os.File
#### Proposal

At the moment, the input and output streams for the REPL are `*os.File`. I propose that these be made `io.Reader` for the input and `io.Writer` for the output.

#### Background

Using `*os.File` constrains the source and sink types unnecessarily. All the current features can be simply reimplemented basing the parameters off `io.Reader` and `io.Writer`.

This would allow virtual file systems to be used, like the one provided [here](https://godoc.org/golang.org/x/tools/godoc/vfs).

#### Workarounds

It is possible to make a pipe of an `io.Reader` to `*os.File`. Similarly this can be done with the output. However, both of this require spinning up a goroutine to pass the reads and writes through the pipe.

kortschak:
Please take a look.

Closes #403.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -6,6 +6,7 @@ import (
 	"go/build"
 	"go/scanner"
 	"go/token"
+	"io"
 	"os"
 	"reflect"
 	"strconv"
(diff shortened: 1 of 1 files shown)


mvertes:
In the same time, I will perform the transition from `Repl` to  `REPL`

PR #406, opened 2019-10-10
feature: add support for custom build tags

### LINKED ISSUE #372: add command line option to set custom build tags
#### Proposal

Add `-tags tag,list` to yaegi command, with the same effect as `go build -tags tag,list`, i.e defining the tags as custom build constraints used in comments such as `// +build !amd64 noasm`

#### Background

Some third party packages are using custom build tags to trigger assembly optimisations, or to avoid `unsafe` package, etc. 

For example, the package `google.golang.org/grpc` uses `purego` tag to avoid  dependency on `unsafe` package. Related to #260 

In another example, `gonum.org/v1/gonum/internal/math32` contains some implementation of Sqrt avoiding assembly dependency using `noasm` constraint. Related to #301.

#### Workarounds

No workaround known

mvertes:
Fixes #372

### DIFF
--- a/interp/doc.go
+++ b/interp/doc.go
@@ -13,14 +13,14 @@ future imports of the session.
 
 A build tag is a line comment that begins
 
-	// yaegi:tags:
+	// yaegi:tags
 
 that lists the build constraints to be satisfied by the further
 imports of source packages.
 
 For example the following custom build tag
 
-	// yaegi:tags: noasm
+	// yaegi:tags noasm
 
 Will ensure that an import of a package will exclude files containing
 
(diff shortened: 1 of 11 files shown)


interp/doc.go
```diff
@@ -13,14 +13,14 @@ future imports of the session.
 
 A build tag is a line comment that begins
 
-	// yaegi-tags:
+	// yaegi:tags:
```
mvertes: remove trailing `:`

interp/doc.go
```diff
@@ -13,14 +13,14 @@ future imports of the session.
...
 
 that lists the build constraints to be satisfied by the further
 imports of source packages.
 
 For example the following custom build tag
 
-	// yaegi-tags: noasm
+	// yaegi:tags: noasm
```
mvertes: remove trailing `:`

mvertes:
Last changes ok to me

PR #432, opened 2019-10-31
fix: automatic conversion of untyped literal float to int

### LINKED ISSUE #428: yaegi incorrectly types 1e2 as a non-integer
The following program `sample.go` fails to run with "illegal operand types for '%' operator":

```
func main() {
	i := 100
	if i%1e2 == 0 {
		println(i)
	}
}
```

Expected result:
```console
$ go run ./sample.go
100
```

Got:
```console
$ yaegi ./sample.go
sample.go:5:5: illegal operand types for '%' operator
```

The code with `1e2` replaced by `100` works as expected.

mvertes:
Fixes #428


kortschak:
I don't think this is right, though I also don't think that it's possible to be correct without a separate typish system to handle untyped constants as Go has.

To demonstrate this:
```
~ $ cat fl.go 
package main

func main() {
	const a = 100.0
	println(a % 100)
}
~ $ go run fl.go 
# command-line-arguments
./fl.go:5:12: invalid operation: operator % not defined on untyped float
~ $ yaegi fl.go 
0
```

kortschak:
Related:
```
~ $ cat fl2.go 
package main

import "fmt"

func main() {
	const a = 1.0 / 3.0
	const b = a * 3
	fmt.Printf("%T %.20[1]f\n", b)
}
~ $ go run fl2.go 
float64 1.00000000000000000000
~ $ yaegi fl2.go 
int %!f(int=00000000000000000000)
```

mvertes:
The type check logic for constants is indeed not yet correct. I saw that after my question on [golang-nuts](https://groups.google.com/forum/#!topic/golang-nuts/tQQSl0LIKkE). I misunderstood the spec on that aspect and will need to revisit.

PR #445, opened 2019-11-15
fix: detect non boolean condition for IF and FOR statements

### LINKED ISSUE #438: yaegi incorrectly accepts non-bool conditional expression in if statement
The following program `cond.go` unexpectedly runs:

```go
package main

import "fmt"

func main() {
	var i int
	if i % 1000000 {
		fmt.Println("oops")
	}
}
```

Expected result:
```console
~ $ go run cond.go 
# command-line-arguments
./cond.go:7:2: non-bool i % 1000000 (type int) used as if condition
```

Got:
```console
~ $ yaegi cond.go
oops
```

mvertes:
Add condition type check in post-processing of CFG for
if and for statements. Add an isBool() helper function
which also accepts boolean channel values.

Fixes #438

### DIFF
--- a/interp/type.go
+++ b/interp/type.go
@@ -935,6 +935,17 @@ func isInterface(t *itype) bool {
 
 func isStruct(t *itype) bool { return t.TypeOf().Kind() == reflect.Struct }
 
+func isBool(t *itype) bool {
+	switch t.TypeOf().Kind() {
+	case reflect.Bool:
+		return true
+	case reflect.Chan:
+		return t.TypeOf().Elem().Kind() == reflect.Bool
+	default:
+		return false
+	}
+}
+
 func isInt(t reflect.Type) bool {
 	switch t.Kind() {
 	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64, reflect.Uint, reflect.Uint8, reflect.Uint16, reflect.Uint32, reflect.Uint64, reflect.Uintptr:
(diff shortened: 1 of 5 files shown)


interp/type.go
```diff
@@ -935,6 +935,17 @@ func isInterface(t *itype) bool {
 
 func isStruct(t *itype) bool { return t.TypeOf().Kind() == reflect.Struct }
 
+func isBool(t *itype) bool {
+	switch t.TypeOf().Kind() {
+	case reflect.Bool:
+		return true
+	case reflect.Chan:
+		return t.TypeOf().Elem().Kind() == reflect.Bool
```
kortschak: Can you explain this part?

mvertes: This is necessary to handle `if` expression on a boolean channel, like:
```go
channel := make(chan bool)
if <-channel {
 ....
}
```

kortschak: Odd. `<-(chan bool)` is `bool`; `chan bool` is `chan bool`. I would love to understand how this is all constructed here.

mvertes: Good question! the type computation of index expression resolving to a bool and function call returning a bool, as performed during CFG in  `nodeType()`,  returns a `bool` type as expected. So aanalysis of a  chan recv expression on a boolean channel should produce the same. I need to  take a closer look, but it may be because i do not take the channel direction in account when performing type analysis on channel expression.

mvertes: The problem was that type from unary operator `<-` was the channel type instead of the channel data type itself. Fixing that removes the need for special processing in `isBool()` , allows other simplifications and probably prevents further bugs. Thanks again for pointing that!

PR #454, opened 2019-12-09
fix: add method checks for interface types

### LINKED ISSUE #452: interface type check false negative
The following program `sample.go` triggers an error:

```go
package main

type T struct{}

func (t *T) Error() string { return "T: error" }

var invalidT = &T{}

func main() {
    var err error
    if err != invalidT {
        println("ok")
    }   
}
```

Expected result:
```console
$ go run ./sample.go
ok
```

Got:
```console
$ yaegi ./sample.go
./sample.go:13:5: mismatched types .error and *main.T
```

mvertes:
Fixes #452


interp/type.go
```diff
@@ -636,6 +636,47 @@ func (t *itype) finalize() (*itype, error) {
 	return t, err
 }
 
+// identical returns true if the given type is identical to the receiver one
+func (t *itype) identical(o *itype) bool {
```
ldez: maybe you can use `equal` or `isEqual` instead of `identical`.

mvertes: renamed `identical` to `equal`

Issue #487, opened 2020-01-11
bug: EvalWithContext makes it impossible to handle panics.


james-lawrence:
The following program `sample.go` triggers a panic:

```go
package main

func main() {
	panic("fail")
}
```

Eval will panic in the callstack of Eval.

EvalWithContext panics on a goroutine. causing a crash.

kortschak:
There are two-ish solutions for this. One is to recover the panic in the goroutine and then send that on a chan to be handled by the `EvalWithContext` finalising switch.

```diff
diff --git a/interp/interp.go b/interp/interp.go
index 8fe3de8..504a44a 100644
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -391,7 +391,14 @@ func (interp *Interpreter) EvalWithContext(ctx context.Context, src string) (ref
        interp.mutex.Unlock()
 
        done := make(chan struct{})
+       panicked := make(chan interface{})
        go func() {
+               defer func() {
+                       r := recover()
+                       if r != nil {
+                               panicked <- r
+                       }
+               }()
                defer close(done)
                v, err = interp.Eval(src)
        }()
@@ -402,6 +409,8 @@ func (interp *Interpreter) EvalWithContext(ctx context.Context, src string) (ref
                return reflect.Value{}, ctx.Err()
        case <-done:
                return v, err
+       case r := <-panicked:
+               panic(r)
        }
 }
 
```

This unfortunately loses the context of the panic, and also has a conceptual defect of conflating the idea that panicking in the interpreted code means a panic in the host program. I think this is wrong on the basis that I wouldn't expect a program's panic to crash my OS or blow up my computer, so why should an interpreted program bring down my host program.

Probably a better approach is to make an `interp.Panic` type that satisfies `error` and is handled for all `Eval` calls, and captures the stack.

This is the diff for that.

```diff
diff --git a/interp/interp.go b/interp/interp.go
index 8fe3de8..4acaa8d 100644
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -11,6 +11,7 @@ import (
        "os"
        "os/signal"
        "reflect"
+       "runtime"
        "strconv"
        "sync"
        "sync/atomic"
@@ -158,6 +159,18 @@ type _error struct {
 
 func (w _error) Error() string { return w.WError() }
 
+// Panic is an error recovered from a panic call in interpreted code.
+type Panic struct {
+       // Value is the recovered value of a call to panic.
+       Value interface{}
+
[...]

Issue #413, opened 2019-10-12
Bug: Eval panics.


james-lawrence:
it shouldn't. it already returns an error. it should recover itself and return that error.
```go
// Eval evaluates Go code represented as a string. It returns a map on
// current interpreted package exported symbols
func (interp *Interpreter) Eval(src string) (res reflect.Value, err error) {
	defer func() {
		recovered := recover()
		if recovered == nil {
			return
		}

		switch cause := recovered.(type) {
		case runtime.Error:
                        panic(cause)
		case error:
                        err = cause
		}
	}()

	// Parse source to AST
	pkgName, root, err := interp.ast(src, interp.Name)
	if err != nil || root == nil {
		return res, err
	}

	if interp.astDot {
		root.astDot(dotX(), interp.Name)
		if interp.noRun {
			return res, err
		}
	}

	// Global type analysis
	revisit, err := interp.gta(root, pkgName)
	if err != nil {
		return res, err
	}
	for _, n := range revisit {
		if _, err = interp.gta(n, pkgName); err != nil {
			return res, err
		}
	}

	// Annotate AST with CFG infos
	initNodes, err := interp.cfg(root)
	if err != nil {
		return res, err
	}

	// Add main to list of functions to run, after all inits
	if m := interp.main(); m != nil {
		initNodes = append(initNodes, m)
	}

	if root.kind != fileStmt {
		// REPL may skip package statement
		setExec(root.start)
	}
	if interp.universe.sym[pkgName] == nil {
		// Make the package visible under a path identical to its name
		interp.srcPkg[pkgName] = interp.scopes[pkgName].sym
		interp.universe.sym[pkgName] = &symbol{kind: pkgSym, typ: &itype{cat: srcPkgT, path: pkgName}}
	}

	if interp.cfgDot {
		root.cfgDot(dotX())
	}

	if interp.noRun {
		return res, err
	}

	// Execute CFG
	if err = genRun(root); err != nil {
		return res, err
	}
	interp.resizeFrame()
	interp.run(root, nil)

	for _, n := range initNodes {
		interp.run(n, interp.frame)
	}
	v := genValue(root)
	res = v(interp.frame)

	// If result is an interpreter node, wrap it in a runtime callable function
	if res.IsValid() {
		if n, ok := res.Interface().(*node); ok {
			res = genFunctionWrapper(n)(interp.frame)
		}
	}

	return res, err
}
```

mvertes:
[...]

PR #508, opened 2020-02-08
interp: weaken panics to errors and return panicked values

kortschak:
Please take a look.

Updates #413.
Updates #487.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -12,6 +12,7 @@ import (
 	"os/signal"
 	"reflect"
 	"runtime"
+	"runtime/debug"
 	"strconv"
 	"sync"
 	"sync/atomic"
@@ -167,6 +168,9 @@ type Panic struct {
 	// Callers is the call stack obtained from the recover call.
 	// It may be used as the parameter to runtime.CallersFrames.
 	Callers []uintptr
+
+	// Stack is the call stack buffer for debug.
+	Stack []byte
 }
 
 // TODO: Capture interpreter stack frames also and remove
@@ -313,7 +317,7 @@ func (interp *Interpreter) Eval(src string) (res reflect.Value, err error) {
 		if r != nil {
 			var pc [64]uintptr // 64 frames should be enough.
 			n := runtime.Callers(1, pc[:])
-			err = Panic{Value: r, Callers: pc[:n]}
+			err = Panic{Value: r, Callers: pc[:n], Stack: debug.Stack()}
 		}
 	}()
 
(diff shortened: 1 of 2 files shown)


interp/interp.go
```diff
@@ -289,10 +305,15 @@ func (interp *Interpreter) main() *node {
 	return nil
 }
 
-// Eval evaluates Go code represented as a string. It returns a map on
-// current interpreted package exported symbols
-func (interp *Interpreter) Eval(src string) (reflect.Value, error) {
-	var res reflect.Value
+func (interp *Interpreter) Eval(src string) (res reflect.Value, err error) {
```
james-lawrence: you lost the method documentation.

kortschak: Eep. Fixed.

mvertes:
Thank you for your efforts. Sorry to bring my response a bit late in the process.

I have to say that  I'm not convinced that recovering panics _inside_ `Eval` is a better practice than recovering it _outside_, i.e. in the host application (if necessary). My rationale is:
- as panic can be recovered outside of `Eval` anyway, it is not to me a showstopper if it is not implemented in yaegi, the user can always achieve its desired behavior at a small cost.
- panics may be of different origins, system problem, bug in yaegi,, script action. etc. I'm not sure that transforming all these in a yaegi error is the right thing to do. In fact I don't know the best thing to do (and can't), and therefore the best (to me) is still to let the panic propagate, and let the user decide based on context.
- The context in panic is a full stack trace, invaluable for debugging and post-mortem (if no recover), and much richer than a custom error. With the current patch, I loose the ability to debug, and forbid the user to debug as well. Or we could generate stack trace part of error info, but in this case, why not just propagate panics ?

mvertes:
I just went through the PR again, and noticed that you indeed capture the info necessary to access the stack traces, sorry for having missed that. Although it doesn't change the conclusion.

kortschak:
I differ in opinion. What you find is that basically all uses of `Eval` without this then must be wrapped in a recover in order to avoid bringing down the application.

kortschak:
Note that all the things that a user can do with recovery they user can do with a type assertion on the returned `error` from `Eval` and then introspection on the `interp.Panic.Value`. With some additional work, the user could even have a stack trace of the interpreted code (which would not be possible at all with the propagated panic).

kortschak:
*ping*

mvertes:
I added a formatted stack buffer (as returned by `runtime/debug.Stack()`) to preserve the previous tracing behaviour, and still be able to recover automatically. `Panic.Callers` could indeed be used to dump a call trace at interpreter level in the future.

kortschak:
Thanks

PR #516, opened 2020-02-13
fix: didn't set values after calculation while there is a fnext branch

dusmart:
I don't know if this method is the right way according to your code style. 
If this is really useful for fixing those bugs, I'm sure that there are lots of similar code blocks that need to be fixed which you can search by keyword `if n.fnext != nil`.

The following program sample.go triggers an unexpected error:
```
package main
type Fromage struct {
	Done bool
}

func main() {
	a := Fromage{
		true,
	}

	println( a.Done && true )
}
```
Expected result:

$ go run ./sample.go
true

Got:

$ yaegi ./sample.go
false

### DIFF
--- a/interp/interp_eval_test.go
+++ b/interp/interp_eval_test.go
@@ -245,6 +245,76 @@ func f() string {
 	}
 }
 
+func TestEvalStruct2(t *testing.T) {
+	i := interp.New(interp.Options{})
+	eval(t, i, `
+type Fromage struct {
+	Done bool
+}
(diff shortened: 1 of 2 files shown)


mvertes:
Thank you for  your report and proposal. Please could you create a separate issue (not in a pull request), so we can track this problem properly. No need to change the content.

Your solution fixes this particular test, but I need more time to assess if it is addressing all possible cases.

Thanks again.

PR #518, opened 2020-02-13
Add option "-e"

xaionaro:
Option ~~"-c"~~ "-e" (updated) allows to set the command to be executed

```
echo '6001d5ff0000000003000000000107000000dcff' | \
	yaegi -e 'import "fmt"; import "os"; import "encoding/hex"; import "io/ioutil"; func main() { in, _ := ioutil.ReadAll(os.Stdin); decoded, _ := hex.DecodeString(string(in)); fmt.Println(string(decoded)) }' 2>/dev/null | \
	hexdump -C
00000000  60 01 d5 ff 00 00 00 00  03 00 00 00 00 01 07 00  |`...............|
00000010  00 00 dc ff 0a                                    |.....|
00000015
```

Of course it is understandable if this PR will be rejected :)

### DIFF
--- a/cmd/yaegi/yaegi.go
+++ b/cmd/yaegi/yaegi.go
@@ -47,8 +47,10 @@ import (
 func main() {
 	var interactive bool
 	var tags string
+	var cmd string
 	flag.BoolVar(&interactive, "i", false, "start an interactive REPL")
 	flag.StringVar(&tags, "tags", "", "set a list of build tags")
+	flag.StringVar(&cmd, "e", "", "set the command to be executed (instead of script or/and shell)")
 	flag.Usage = func() {
 		fmt.Println("Usage:", os.Args[0], "[options] [script] [args]")
 		fmt.Println("Options:")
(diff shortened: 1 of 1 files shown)


xaionaro:
Nah, too dirty...

mvertes:
I love this PR, except I would prefer `-e` for eval :smile:

xaionaro:
@mvertes : Yeah, sure. But I need to remove `strings.Replace` (it's a dirty hack), though. I will fix it tomorrow. Thank you for the interest :)

xaionaro:
I did not find a nice quick way to implement this. Let it be as it is :)

It would be nice to implement an auto-importer, so instead of:
```
import "fmt"; import "os"; import "encoding/hex"; import "io/ioutil"; func main() { in, _ := ioutil.ReadAll(os.Stdin); decoded, _ := hex.DecodeString(string(in)); fmt.Println(string(decoded)) }
```
it will be enough to write just:
```
in, _ := ioutil.ReadAll(os.Stdin); decoded, _ := hex.DecodeString(string(in)); fmt.Println(string(decoded))
```

ldez:
@xaionaro has your PR comes from a GitHub Organization, neither us nor our bot can manage conflicts on your PR.

So could you fix the conflict, thank you.

ldez:
Merged in d29b0a48ffb20bb0bba5981e9125fa28b5fd5d62

PR #545, opened 2020-03-12
fix: resolve receiver for binary methods on non interface types

### LINKED ISSUE #542: inherit struct function bug

```go
package main
type S struct {
	http.Client
}
func main() {
	var s S
	_, err := s.Get("url")
	fmt.Println(err)
	return
}
```

Expected result:
```console
$ go run ./sample.go
Get "url": unsupported protocol scheme ""
```

Got:
```console
$ yaegi ./sample.go
 reflect.Value.Convert: value of type string cannot be converted to type *http.Client
```

mvertes:
Fixes #542

### DIFF
--- a/_test/struct36.go
+++ b/_test/struct36.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"fmt"
+	"net/http"
+)
+
+type S struct {
+	http.Client
+}
+
+func main() {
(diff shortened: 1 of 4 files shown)


mvertes:
The problem is fixed, but I saw that the solution could be simplified a bit.

PR #578, opened 2020-04-14
fix: correct control flow graph for select blocks

### LINKED ISSUE #577: select on channel struct field

```go
ackage main
  
type T struct {
    c1 chan string
}

func main() {
    t := &T{}
    t.c1 = make(chan string)

    go func() { t.c1 <- "done" }()

    select {
    case msg1 := <-t.c1:
        println("received from c1:", msg1)
    }
    println("Bye")
}
```

Expected result:
```console
$ go run ./sample.go
received from c1: done
Bye
```

Got:
```console
$ yaegi ./sample.go
./sample.go:15:32: undefined: msg1
```

mvertes:
Receive channel in comm clause of select statements were only supported
for channel variables. Now channel init actions, for example to access
a channel struct field, are properly chained prior to invoke the select
operation. In addition, verify that the receive operation is on a
channel type data.

Fixes #577.

### DIFF
--- a/_test/chan9.go
+++ b/_test/chan9.go
@@ -1,0 +1,20 @@
+package main
+
+type Channel chan string
+
+type T struct {
+	Channel
+}
+
+func send(c Channel) { c <- "ping" }
+
+func main() {
+	t := &T{}
+	t.Channel = make(Channel)
+	go send(t.Channel)
+	msg := <-t.Channel
+	println(msg)
+}
+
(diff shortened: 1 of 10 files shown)


mvertes:
Another test case and fix is missing for that.

mvertes:
Tested with an empty body in comm clause which is causing an infinite loop. Another fix is needed.

PR #632, opened 2020-05-06
fix: compute numeral constant on typed objects

### LINKED ISSUE #631: Defining typed constant numbers

```go
package main

type Kind int

const (
    None Kind = 0
    Left Kind = 1 << iota
    Right
    Both Kind = Left | Right
)

func main() {
    println(None, Left, Right, Both)
}
```

Expected result:
```console
$ go run ./sample.go
0 2 4 6
```

Got:
```console
$ yaegi ./sample.go
sample.go:7:2: constant definition loop
```

mvertes:
Remove the spurious constraint that a constant must be untyped
when testing for arithmetic properties.

Fixes #631.

### DIFF
--- a/_test/const12.go
+++ b/_test/const12.go
@@ -1,0 +1,17 @@
+package main
+
+type Kind int
+
+const (
+	None Kind = 0
+	Left Kind = 1 << iota
+	Right
+	Both Kind = Left | Right
+)
+
+func main() {
+	println(None, Left, Right, Both)
+}
+
+// Output:
+// 0 2 4 6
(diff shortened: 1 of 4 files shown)


_test/const12.go
```diff
@@ -0,0 +1,17 @@
+package main
```
mpl, line 1: I think that instead of this test, we should rather add a simple unit test for isNatural that checks that it returns true for a typed const as well. I can try to do it if you want?

mvertes:
I agree with your suggestion to apply this PR (with additional test on `isNatural`), and keep #631 open for now.

PR #643, opened 2020-05-19
fix: use interface wrappers to expose interface values to runtime

### LINKED ISSUE #630: Struct to Stringer var assignment panics

```go
package main

import (
	"fmt"
)

type foo struct {
	bar string
}

func (f foo) String() string {
	return f.bar
}

func main() {
	var f fmt.Stringer = foo{bar: "bar"}
	fmt.Println(f)
}
```

Expected result:
```console
$ go run ./sample.go
bar
```

Got:
```console
$ yaegi ./sample.go
./sample.go:16:8: panic
reflect.Set: value of type struct { Xbar string } is not assignable to type fmt.Stringer
goroutine 1 [running]:
runtime/debug.Stack(0x1, 0xc000148400, 0x40)
        /usr/local/Cellar/go/1.14.2_1/libexec/src/runtime/debug/stack.go:24 +0x9d
github.com/containous/yaegi/interp.(*Interpreter).Eval.func1(0xc0001a7e00)
        /Users/berng/go/src/github.com/containous/yaegi/interp/interp.go:320 +0xbc
panic(0x1896ea0, 0xc000192a70)
        /usr/local/Cellar/go/1.14.2_1/libexec/src/runtime/panic.go:969 +0x166
github.com/containous/yaegi/interp.runCfg.func1(0xc0001357c0, 0xc000300a00)
        /Users/berng/go/src/github.com/containous/yaegi/interp/run.go:113 +0x20c
panic(0x1896ea0, 0xc000192a70)
        /usr/local/Cellar/go/1.14.2_1/libexec/src/runtime/panic.go:969 +0x166
reflect.Value.assignTo(0xc0001207e0, 0xc000192a00, 0x199, 0x19c9d58, 0xb, 0x18edb60, 0xc000192a40, 0xc0002da8a0, 0xc000192860, 0xc0001425a0)
        /usr/local/Cellar/go/1.14.2_1/libexec/src/reflect/value.go:2403 +0x426
reflect.Value.Set(0x18edb60, 0xc000192a40, 0x194, 0xc0001207e0, 0xc000192a00, 0x199)
        /usr/local/Cellar/go/1.14.2_1/libexec/src/reflect/value.go:1532 +0xbd
github.com/containous/yaegi/interp.doCompositeSparse.func1(0xc0001357c0, 0xc000043f80)
        /Users/berng/go/src/github.com/containous/yaegi/interp/run.go:1871 +0x264
github.com/containous/yaegi/interp.runCfg(0xc000300a00, 0xc0001357c0)
        /Users/berng/go/src/github.com/containous/yaegi/interp/run.go:119 +0x6d
github.com/containous/yaegi/interp.(*Interpreter).run(0xc000001980, 0xc000300000, 0xc000135720)
        /Users/berng/go/src/github.com/containous/yaegi/interp/run.go:97 +0x239
github.com/containous/yaegi/interp.(*Interpreter).Eval(0xc000001980, 0xc0002fe000, 0xb8, 0x0, 0x0, 0x0, 0x0, 0x0)
        /Users/berng/go/src/github.com/containous/yaegi/interp/interp.go:388 +0x41f
main.main()
        /Users/berng/go/src/github.com/containous/yaegi/cmd/yaegi/yaegi.go:139 +0x554
```

mvertes:
If a value is assigned to, or returned as, a binary interface,
then use the interface wrapper generator to convert the value
accordingly.

Fixes #630.

### DIFF
--- a/_test/interface40.go
+++ b/_test/interface40.go
@@ -10,12 +10,12 @@ func (f foo) String() string {
 	return "Hello from " + f.bar
 }
 
-func NewFoo(s string) fmt.Stringer {
+func Foo(s string) fmt.Stringer {
 	return foo{s}
 }
 
 func main() {
-	f := NewFoo("bar")
+	f := Foo("bar")
 	fmt.Println(f)
 }
 
--- a/_test/interface40.go
+++ b/_test/interface40.go
@@ -1,0 +1,23 @@
+package main
+
+import "fmt"
+
+type foo struct {
+	bar string
+}
+
+func (f foo) String() string {
+	return "Hello from " + f.bar
+}
+
+func NewFoo(s string) fmt.Stringer {
+	return foo{s}
+}
+
+func main() {
+	f := NewFoo("bar")
+	fmt.Println(f)
+}
+
+// Output:
+// Hello from bar
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -507,9 +507,13 @@ func (interp *Interpreter) cfg(root *node, pkgID string) ([]*node, error) {
 					src.findex = dest.findex // Set recv address to LHS
 					dest.typ = src.typ
 				case n.action == aAssign && src.action == aCompositeLit:
-					n.gen = nop
-					src.findex = dest.findex
-					src.level = level
+					if !(dest.typ.cat == valueT && dest.typ.rtype.Kind() == reflect.Interface) {
+						// Output literal composite to destination (avoiding assign operation)
+						// only if there is no need to wrap to an interface.
+						n.gen = nop
+						src.findex = dest.findex
+						src.level = level
+					}
 				case src.kind == basicLit:
 					// TODO: perform constant folding and propagation here.
 					switch {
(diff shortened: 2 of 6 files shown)


_test/interface40.go
```diff
@@ -0,0 +1,23 @@
...
+
+type foo struct {
+	bar string
+}
+
+func (f foo) String() string {
+	return "Hello from " + f.bar
+}
+
+func NewFoo(s string) fmt.Stringer {
```
mpl, line 13: By convention, a function prefixed with "New" usually returns a pointer. So the "problem" here is two-fold: 1) the result in the signature is an interface, not a pointer, and 2) a foo is returned, not a pointer to a foo.
I would suggest s/NewFoo/Foo/ for example.

mvertes: ok, done

interp/cfg.go
```diff
@@ -507,9 +507,13 @@ func (interp *Interpreter) cfg(root *node, pkgID string) ([]*node, error) {
 					src.findex = dest.findex // Set recv address to LHS
 					dest.typ = src.typ
 				case n.action == aAssign && src.action == aCompositeLit:
-					n.gen = nop
-					src.findex = dest.findex
-					src.level = level
+					if !(dest.typ.cat == valueT && dest.typ.rtype.Kind() == reflect.Interface) {
```
mpl, line 510: A common practice in Go is to "return early" as much as possible, to avoid extra indentation, and hence improve readability. So you could invert that if, and unindent the whole block, i.e.:
```
if (dest.typ.cat == valueT && dest.typ.rtype.Kind() == reflect.Interface) {
    break
}
n.gen = nop
...
```

PR #650, opened 2020-05-26
interp: extend dot debugging

mpl:
As dotty does not seem to be readily available on MacOS, this PR extends
the ability to debug with dot, through the new YAEGI_DOT_CMD env var,
which allows one to specify the command to process the generated dot
graphs.

For example, to print the AST to svg:

YAEGI_AST_DOT=1 YAEGI_DOT_CMD='dot -Tsvg -ofoo.svg' yaegi ./foo.go

If any of YAEGI_AST_DOT or YAEGI_CFG_DOT is set, but YAEGI_DOT_CMD is
not defined, the default is to write to a .dot file next to the Go
source file.


mpl:
ah I've just realized if both YAEGI_AST_DOT and YAEGI_CFG_DOT they would overwrite each other as they would write to the same default file name. I need to add a differentiator in the filename.

PR #657, opened 2020-05-27
fix: finish support of type assertions which was incomplete

### LINKED ISSUE #640: value of type struct is not assignable to type struct
Version:
v0.8.5 (cdf46224212112b154d5a536e5f3916abee50bd8)

```go
package main

import (
        "fmt"
)

type T1 struct{}

func (t1 T1) f() {
        fmt.Println("T1.f()")
}

func (t1 T1) g() {
        fmt.Println("T1.g()")
}

type T2 struct {
        T1
}

func (t2 T2) f() {
        fmt.Println("T2.f()")
}

type I interface {
        f()
}

func printType(i I) {
        if t1, ok := i.(T1); ok {
                t1.f()
                t1.g()
        }

        if t2, ok := i.(T2); ok {
                t2.f()
                t2.g()
        }
}

func main() {
        printType(T1{})
        printType(T2{})
}
```

Expected result:
```console
$ go run ./sample.go
 T1.f()
 T1.g()
 T2.f()
 T1.g()
```

Got:
```console
$ yaegi ./sample.go
T1.f()
T1.g()
/home/zimash/go/sample.go:30:15: panic
/home/zimash/go/sample.go:42:12: panic
reflect.Set: value of type struct {} is not assignable to type struct { T1 struct {} }
goroutine 1 [running]:
runtime/debug.Stack(0x1, 0xc000164600, 0x40)
        /usr/local/go/src/runtime/debug/stack.go:24 +0x9d
github.com/containous/yaegi/interp.(*Interpreter).Eval.func1(0xc00009be00)
        /home/zimash/goos/src/github.com/containous/yaegi/interp/interp.go:320 +0xbc
panic(0xc97340, 0xc0001c1470)
        /usr/local/go/src/runtime/panic.go:967 +0x15d
github.com/containous/yaegi/interp.runCfg.func1(0xc00014f680, 0xc000324300)
        /home/zimash/goos/src/github.com/containous/yaegi/interp/run.go:113 +0x20c
panic(0xc97340, 0xc0001c1470)
        /usr/local/go/src/runtime/panic.go:967 +0x15d
github.com/containous/yaegi/interp.runCfg.func1(0xc00014f720, 0xc000321800)
        /home/zimash/goos/src/github.com/containous/yaegi/interp/run.go:113 +0x20c
panic(0xc97340, 0xc0001c1470)
        /usr/local/go/src/runtime/panic.go:967 +0x15d
reflect.Value.assignTo(0xcc6de0, 0x15928e0, 0x199, 0xdc9608, 0xb, 0xc0001367e0, 0x0, 0xc00009b9a0, 0xc00014f9a0, 0x0)
        /usr/local/go/src/reflect/value.go:2403 +0x426
reflect.Value.Set(0xc0001367e0, 0x15928e0, 0x199, 0xcc6de0, 0x15928e0, 0x199)
        /usr/local/go/src/reflect/value.go:1532 +0xbd
github.com/containous/yaegi/interp.typeAssert2.func3(0xc00014f720, 0xc0002faed0)
        /home/zimash/goos/src/github.com/containous/yaegi/interp/run.go:214 +0x102
github.com/containous/yaegi/interp.runCfg(0xc000321800, 0xc00014f720)
        /home/zimash/goos/src/github.com/containous/yaegi/interp/run.go:119 +0x6d
github.com/containous/yaegi/interp.call.func7(0xc00014f680, 0xc0003309a0)
        /home/zimash/goos/src/github.com/containous/yaegi/interp/run.go:840 +0xbf1
github.com/containous/yaegi/interp.runCfg(0xc000324300, 0xc00014f680)
        /home/zimash/goos/src/github.com/containous/yaegi/interp/run.go:119 +0x6d
github.com/containous/yaegi/interp.(*Interpreter).run(0xc000102a80, 0xc000323a00, 0xc00014f5e0)
        /home/zimash/goos/src/github.com/containous/yaegi/interp/run.go:97 +0x239
github.com/containous/yaegi/interp.(*Interpreter).Eval(0xc000102a80, 0xc000318000, 0x199, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/zimash/goos/src/github.com/containous/yaegi/interp/interp.go:388 +0x41f
main.main()
        /home/zimash/goos/src/github.com/containous/yaegi/cmd/yaegi/yaegi.go:139 +0x554
```

mvertes:
TypeAssert was optimistically returning ok without verifying that
value could be converted to the required interface (in case of
type assert of an interface type), or not checking the type in
all conditions.  There is now a working `implements` method for
`itype`.

The implementation also relies on internal type checks
performed by the go runtime when assertions operate
on non-emulated types, whenever possible.

A couple of calls to panic() have been added to match
the specification (assert with no status). It still can be difficult to
distinguish between a panic due to wrong user code,
or a bug in yaegi. This will be improved later.

Fixes #640.


interp/type.go
```diff
@@ -841,11 +841,19 @@ func (t *itype) methods() methodSet {
 				res[m.ident] = m.typ.TypeOf().String()
 			}
 		}
-	default:
-		for _, m := range t.method {
-			res[m.ident] = m.typ.TypeOf().String()
+		for k, v := range t.val.methods() {
```
mpl, line 844: Reminder as per chat: this loop seems to have the counter-effect of the filter above. Also, we need to check whether L839 and L844 have different effects, to pick the right one that we want to keep.

mvertes: I removed this code block which is useless. Thanks for noticing.

interp/run.go
```diff
@@ -124,31 +124,35 @@ func typeAssertStatus(n *node) {
 	c0, c1 := n.child[0], n.child[1]
 	value := genValue(c0)              // input value
 	value1 := genValue(n.anc.child[1]) // returned status
-	typ := c1.typ.rtype                // type to assert
+	typ := c1.typ                      // type to assert
 	next := getExec(n.tnext)
 
 	switch {
-	case c0.typ.cat == valueT:
+	case isInterfaceSrc(c1.typ):
```
mpl, line 131:
```suggestion
	case isInterfaceSrc(typ):
```

interp/run.go
```diff
@@ -124,31 +124,35 @@ func typeAssertStatus(n *node) {
...
+		n.exec = func(f *frame) bltn {
+			v, ok := value(f).Interface().(valueInterface)
+			value1(f).SetBool(ok && v.node.typ.implements(typ))
+			return next
+		}
+	case isInterface(c1.typ):
```
mpl, line 137:
```suggestion
	case isInterface(typ):
```

interp/type.go
```diff
@@ -835,17 +835,19 @@ func (t *itype) methods() methodSet {
...
 		}
-	default:
-		for _, m := range t.method {
-			res[m.ident] = m.typ.TypeOf().String()
+	case structT:
+		for _, f := range t.field {
+			for k, v := range f.typ.methods() {
+				res[k] = v
+			}
 		}
```
mpl, line 846:
```suggestion
		}
		for _, m := range t.method {
			res[m.ident] = m.typ.TypeOf().String()
		}		
```

interp/type.go
```diff
@@ -835,17 +835,19 @@ func (t *itype) methods() methodSet {
...
 	}
+	for _, m := range t.method {
+		res[m.ident] = m.typ.TypeOf().String()
+	}
```
mpl, line 850:
```suggestion
```
If we move these lines according to my suggestions, it seems like all the tests in ./interp still pass.
So I don't know if my suggestion is correct (because I don't fully understand the whole case), or if we are missing tests, but I think if my suggestion is correct, then we should apply it, because it makes it clear that this for loop is only needed in the struct case.

interp/run.go
mvertes: Instead, to make clear that we test the type to assert, and not the input type, we could keep using `c1.type` everywhere except in the closure (spare an indirection for performance). A comment will be added to specify that `c0` contains the input value and `c1.typ` the type to assert.

interp/type.go
mvertes: The following test demonstrates the need to keep this block as is.

```go
package main
  
type T int

func (t T) Error() string { return "T: error" }

var invalidT T

func main() {
    var err error
    if err != invalidT {
        println("ok")
    }
}

// Output:
// ok
```

The block handles the method set for arbitrary types, in addition of specific actions for pointers, interfaces, values, and struct. Here the example defines an alias `int` which satisifies the `error` interface.

Note that it triggers a panic in the `notEqual()` operator, but it is outside the scope of this PR. By moving the block as you propose we fail in a `mismatch types` error which shouyld not happen.

P.S. Just filing a new issue for the above test.

interp/type.go
mvertes: Declined, after analysis, it would not return the correct set of methods for an arbitrary type, see comment on next.

PR #658, opened 2020-05-27
fix: compose recursion

nrwiersma:
This PR fixes the case where a recursive struct is assigned on struct creation by adding recursion handling the sparse composite creation.

### DIFF
--- a/_test/struct46.go
+++ b/_test/struct46.go
@@ -1,0 +1,19 @@
+package main
+
+import "fmt"
+
+type A struct {
+	B string
+	C D
+}
+
+type D struct {
+	E *A
+}
+
+func main() {
+	a := &A{B: "b"}
+	a.C = D{E: a}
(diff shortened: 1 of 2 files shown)


_test/struct46.go
```diff
@@ -0,0 +1,19 @@
...
+type D struct {
+	E *A
+}
+
+func main() {
+	a := &A{B: "b"}
+	a.C = D{E: a}
+	fmt.Println(a.C.E.B)
+}
+
```
mvertes, line 19:
```suggestion

// Output:
// b

```

mvertes: Otherwise the test is skipped with message: `has no comment 'Output:' or 'Error:'` when `go test`.

nrwiersma: Yeah, forgot to put it back. Thanks

PR #677, opened 2020-06-03
fix: handle printf changing nil type

### LINKED ISSUE #659: nil from recover function could not be compared properly
The following program `sample.go` has an unexpected result:

```go
package main

import "fmt"

func div(a, b int) (result int) {
	defer func() {
		r := recover()

		fmt.Printf("r = %#v\n", r)

		if r != nil {
			result = 0
		}
	}()

	return a / b
}

func main() {
	println(div(30, 2))

}

```

Expected result:
```console
$ yaegi ./sample.go

r = <nil>
15
```

Got:
```console
$ yaegi ./sample.go
r = <nil>
0
```

So the nil value r in the defer function is not real "nil" value?

nrwiersma:
When running a function on `r` it was modifying the value, making `isNil` and `isNotNil` no longer match properly.

Fixes #659


_test/recover3.go
```diff
@@ -0,0 +1,22 @@
...
+	}
+	if r != nil {
+		println("exception")
+	}
+}
+
+// Output:
+// hello
+// <nil>
+// world
```
mvertes, line 22: Better to have a `<newline>` char at the end of files

_test/recover4.go
```diff
@@ -0,0 +1,25 @@
...
+	return a / b
+}
+
+func main() {
+	println(div(30, 2))
+}
+
+// Output:
+// r = <nil>
+// 15
```
mvertes, line 25: Please add a `<newline>` here too.

PR #684, opened 2020-06-05
fix: type assertion of non-matching types

### LINKED ISSUE #682: selector error after a type assertion

```go
package main

import "reflect"

type I interface {
    Foo()
}
    
type T struct {
    Name string
}

func (t T) Foo() { println("In Foo", t.Name) }

func main() {
    t := T{"test"} 
    v := reflect.ValueOf(t)
    i := v.Interface().(I)
    i.Foo()
}   
```

Expected result:
```console
$ go run ./sample.go
In Foo test
```

Got:
```console
$ yaegi ./sample.go
./sample.go:19:2: undefined field or method: Foo
```

This issue is a blocker fot #652.

nrwiersma:
Type assertions need to do type validations before setting the values. Case 2 has been left intentionally and will make up its own PR, as I have no idea how to do the type assertion on it yet.

Fixes #682

### DIFF
--- a/_test/type23.go
+++ b/_test/type23.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"net/http"
+	"net/http/httptest"
+)
+
+func main() {
+	var v1 interface{} = 1
+	var v2 interface{}
+	var v3 http.ResponseWriter = httptest.NewRecorder()
+
+	if r1, ok := v1.(string); ok {
+		_ = r1
+		println("unexpected")
+	}
+	if _, ok := v1.(string); ok {
+		println("unexpected")
+	}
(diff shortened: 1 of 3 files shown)


nrwiersma:
Realised I need to check implements and assignable in the type check, and probably handle convertible as well.

mvertes:
This PR is impacted by a previous PR #657 on type assertions. Could you rebase it ?

Also note that there may be some cases not working until we support interface composition (I'm working on it right now).

Thanks

nrwiersma:
I am aware I dont not cover all the cases. I recon we will need to rethink a lot at the end of your work, so I am just ignoring those cases for now. It still moves us a little further forward on type assertions though.

PR #688, opened 2020-06-11
feature: refactor numeric constants and detect overflow

### LINKED ISSUE #448: large untyped constant truncated to zero
The following program `sample.go` outputs the wrong value:
```go
package main
import "fmt"
func main() {
	const huge = 1 << 100
	const large = huge>>38
	fmt.Println(int64(large))
}
```

Expected result:
```console
$ go run ./sample.go
4611686018427387904
```

Got:
```console
$ yaegi ./sample.go
0
```
It seems that untyped constants are truncated if they overflow int64, instead of being arbitrary precision (see https://golang.org/ref/spec#Constants)

nrwiersma:
Go numeric constants can be arbitrarily large until to point when they are used, at which time they must be converted to the actual type, and overflow detected. This is now implemented.

Fixes #448


nrwiersma:
The `goexports` has been fixed in branch: https://github.com/nrwiersma/yaegi/tree/goexport-consts. It will be rebased and opened after this merges.

PR #693, opened 2020-06-13
feature: unsafe type conversion

nrwiersma:
This implements the first step of supporting unsafe. Support for the other unsafe functions still needs to be added.

### DIFF
--- a/cmd/yaegi/yaegi.go
+++ b/cmd/yaegi/yaegi.go
@@ -91,6 +91,7 @@ import (
 
 	"github.com/containous/yaegi/interp"
 	"github.com/containous/yaegi/stdlib"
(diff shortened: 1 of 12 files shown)


cmd/yaegi/yaegi.go
```diff
@@ -112,6 +113,7 @@ func main() {
 	i := interp.New(interp.Options{GoPath: build.Default.GOPATH, BuildTags: strings.Split(tags, ",")})
 	i.Use(stdlib.Symbols)
 	i.Use(interp.Symbols)
+	i.Use(unsafe.Symbols)
```
mvertes, line 116: I would like to make this line conditional to a command line option `-unsafe`.

nrwiersma: great idea

PR #699, opened 2020-06-17
interp: force root scope for binPkgT selector

mpl:


mvertes:
Hi @mpl, could you resubmit this PR, now that #698 has been merged,  so we see only the changes regarding the firx bin package name collision ?

Otherwise, it looks good to me.

Thanks.

PR #702, opened 2020-06-19
Handle Receive and Send channels

### LINKED ISSUE #701: Can't assign a returned chan from a bin call to a read only chan

```go
package main

import (
	"time"
)

func main() {
	var tick <-chan time.Time = time.Tick(time.Millisecond)
	_ = tick
       println("success")
}

// Output:
// success

```

Expected result:
```console
$ go run ./sample.go
success
```

Got:
```console
$ yaegi ./sample.go
0xc0001114a0
case14/main.go:8:30: panic
reflect.Set: value of type <-chan time.Time is not assignable to type chan time.Time
goroutine 1 [running]:
runtime/debug.Stack(0x1, 0xc000152400, 0x40)
	/usr/lib/go/src/runtime/debug/stack.go:24 +0x9d
github.com/containous/yaegi/interp.(*Interpreter).Eval.func1(0xc0000c3df0)
	/home/jlevesy/code/src/github.com/containous/yaegi/interp/interp.go:330 +0xbc
panic(0xc9b420, 0xc0001ac790)
	/usr/lib/go/src/runtime/panic.go:969 +0x166
github.com/containous/yaegi/interp.runCfg.func1(0xc0001397c0, 0xc000145700)
	/home/jlevesy/code/src/github.com/containous/yaegi/interp/run.go:114 +0x20c
panic(0xc9b420, 0xc0001ac790)
	/usr/lib/go/src/runtime/panic.go:969 +0x166
reflect.Value.assignTo(0xc94de0, 0xc0001ac788, 0x92, 0xdce02f, 0xb, 0xc955a0, 0x0, 0xc000111520, 0x1, 0x1)
	/usr/lib/go/src/reflect/value.go:2403 +0x426
reflect.Value.Set(0xc955a0, 0xc00011c720, 0x192, 0xc94de0, 0xc0001ac788, 0x92)
	/usr/lib/go/src/reflect/value.go:1532 +0xbd
github.com/containous/yaegi/interp.callBin.func8(0xc0001397c0, 0xc00011e8a0)
	/home/jlevesy/code/src/github.com/containous/yaegi/interp/run.go:1047 +0x2df
github.com/containous/yaegi/interp.runCfg(0xc000145700, 0xc0001397c0)
	/home/jlevesy/code/src/github.com/containous/yaegi/interp/run.go:120 +0x6d
github.com/containous/yaegi/interp.(*Interpreter).run(0xc0001b61a0, 0xc000144800, 0xc000139720)
	/home/jlevesy/code/src/github.com/containous/yaegi/interp/run.go:98 +0x239
github.com/containous/yaegi/interp.(*Interpreter).Eval(0xc0001b61a0, 0xc00030a000, 0xa1, 0x0, 0x0, 0x0, 0x0, 0x0)
	/home/jlevesy/code/src/github.com/containous/yaegi/interp/interp.go:406 +0x52b
main.main()
	/home/jlevesy/code/src/github.com/containous/yaegi/cmd/yaegi/yaegi.go:143 +0x553
```

jlevesy:
## What does this PR do ?

This PR makes yaegi aware of receive and send types. 

It fixes  #701 which was caused by Yaegi considering [all channels to be bidirectional](https://github.com/containous/yaegi/blob/8365f687e784cb07febc382f68e891a90c00ad0d/interp/type.go#L1118).
In case of a binary call like `time.Tick`, the returned channel can actually be a receive channel and in that case we will try to assign it to a bidirectional channel, causing a panic.

To fix that, I made the AST aware of two new kinds `chanTypeRecv` and `chanTypeSend` then added two new type categories `chanSendT` and `chanRecvT`, so we can instantiate the channel values with the right direction.

Hope this helps.

Thanks @nrwiersma for the help !

### DIFF
--- a/interp/ast.go
+++ b/interp/ast.go
@@ -31,6 +31,8 @@ const (
 	caseBody
 	caseClause
 	chanType
+	chanTypeSend
+	chanTypeRecv
 	commClause
 	commClauseDefault
 	compositeLitExpr
@@ -109,6 +111,8 @@ var kinds = [...]string{
 	caseBody:          "caseBody",
 	caseClause:        "caseClause",
 	chanType:          "chanType",
+	chanTypeSend:      "chanTypeSend",
+	chanTypeRecv:      "chanTypeRecv",
 	commClause:        "commClause",
 	commClauseDefault: "commClauseDefault",
 	compositeLitExpr:  "compositeLitExpr",
@@ -565,7 +569,14 @@ func (interp *Interpreter) ast(src, name string) (string, *node, error) {
 			st.push(addChild(&root, anc, pos, caseClause, aCase), nod)
 
 		case *ast.ChanType:
-			st.push(addChild(&root, anc, pos, chanType, aNop), nod)
+			switch a.Dir {
+			case ast.SEND | ast.RECV:
+				st.push(addChild(&root, anc, pos, chanType, aNop), nod)
+			case ast.SEND:
+				st.push(addChild(&root, anc, pos, chanTypeSend, aNop), nod)
+			case ast.RECV:
+				st.push(addChild(&root, anc, pos, chanTypeRecv, aNop), nod)
+			}
 
 		case *ast.CommClause:
 			kind := commClause
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -402,7 +402,7 @@ func (interp *Interpreter) cfg(root *node, pkgID string) ([]*node, error) {
 				}
 			}
 
-		case arrayType, basicLit, chanType, funcType, interfaceType, mapType, structType:
+		case arrayType, basicLit, chanType, chanTypeRecv, chanTypeSend, funcType, interfaceType, mapType, structType:
 			n.typ, err = nodeType(interp, sc, n)
 			return false
 		}
@@ -1892,7 +1892,7 @@ func isBinType(v reflect.Value) bool { return v.IsValid() && v.Kind() == reflect
 // isType returns true if node refers to a type definition, false otherwise.
 func (n *node) isType(sc *scope) bool {
 	switch n.kind {
-	case arrayType, chanType, funcType, interfaceType, mapType, structType:
+	case arrayType, chanType, chanTypeRecv, chanTypeSend, funcType, interfaceType, mapType, structType:
 		return true
 	case parenExpr, starExpr:
 		if len(n.child) == 1 {
@@ -1920,7 +1920,7 @@ func wireChild(n *node) {
 	// Set start node, in subtree (propagated to ancestors by post-order processing)
 	for _, child := range n.child {
 		switch child.kind {
-		case arrayType, chanType, funcDecl, importDecl, mapType, basicLit, identExpr, typeDecl:
+		case arrayType, chanType, chanTypeRecv, chanTypeSend, funcDecl, importDecl, mapType, basicLit, identExpr, typeDecl:
 			continue
 		default:
 			n.start = child.start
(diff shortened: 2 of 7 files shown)


interp/ast.go
```diff
@@ -565,7 +569,14 @@ func (interp *Interpreter) ast(src, name string) (string, *node, error) {
 			st.push(addChild(&root, anc, pos, caseClause, aCase), nod)
 
 		case *ast.ChanType:
-			st.push(addChild(&root, anc, pos, chanType, aNop), nod)
+			switch a.Dir {
+			case ast.SEND | ast.RECV:
```
nrwiersma, line 573: nitpick: shouldnt this be the default case. It might be better to try work out why it is bidirectional than to work out why it isnt added.

jlevesy: Would it be possible to have chan that is neither send or receive ? :thinking:

nrwiersma: I dont think so, not much of a channel if it cannot send or receive.

jlevesy:
This patch seems to fix https://github.com/containous/yaegi/pull/692#pullrequestreview-433058817 as well :)

interp/scope.go
```diff
@@ -144,7 +144,7 @@ func (s *scope) rangeChanType(n *node) *itype {
 		return nil
 	}
 	switch {
-	case c.typ.cat == chanT:
+	case c.typ.cat == chanT, c.typ.cat == chanRecvT:
```
mvertes, line 147: Range over send channel is possible and should be also addressed here.

jlevesy: :thinking:  I'm not sure it is supported by go itself.

```
(41 ms) yaegi ⍉ λ cat case15/main.go 
package main

func main() {
	var c chan<- struct{} = make(chan struct{})

	for _ = range c {
		println("success")
		return
	}
}

// Output:
// success
(4 ms) yaegi λ go run case15/main.go
# command-line-arguments
case15/main.go:6:10: invalid operation: range c (receive from send-only type chan<- struct {})
```

So what you're expecting is to have yaegi fail with a clean error message here, isn't it ?

Thanks !

mvertes: Sorry, I was wrong. Range is not permitted by go over send channels. It's for receiving only. So yes, catching use of range over a send channel should be nice.

Thanks

interp/cfg.go
```diff
@@ -69,7 +69,12 @@ func (interp *Interpreter) cfg(root *node, pkgID string) ([]*node, error) {
 				// type of array like value is not yet known. This could be fixed in ast structure
 				// by setting array/map node as 1st child of ForRangeStmt instead of 3rd child of
 				// RangeStmt. The following workaround is less elegant but ok.
-				if t := sc.rangeChanType(n.anc); t != nil {
+				var t *itype
```
mvertes, line 72: It would be better to test that the channel type is not a `chanSendT` early and break with error, rather than modify `rangeChanType`.

Also do not forget about the possibility of a binary channel (`typ.cat == valueT`) send-only: `typ.TypeOf().Kind() == reflect.Chan && typ.TypeOf().ChanDir == reflect.SendDir`. And also the fact that the type could be aliased (`typ.cat == aliasT`).

For all those reasons, it's better to have an helper `isSendChan` method on type, and  test early as proposed above. You can find other similar helper in type.go, i.e. `isChan` or `isInterface`, etc.

jlevesy: Alright, will do, thanks !

interp/type.go
```diff
@@ -1320,7 +1320,21 @@ func chanElement(t *itype) *itype {
...
+		return true
+	case t.cat == valueT && t.TypeOf().Kind() == reflect.Chan && t.TypeOf().ChanDir() == reflect.SendDir:
+		return true
+	case t.cat == aliasT:
+		return isSendChan(t.val)
+	default:
+		return false
+	}
+}
+
```
mvertes, line 1336:
```suggestion
func isSendChan(t *itype) bool { rt := t.TypeOf(); return rt.Kind() == reflect.Chan && rt.ChanDir() == reflect.SendDir }
```
is sufficient to encompass all possibilities.

PR #703, opened 2020-06-19
fix: array recursion

nrwiersma:
This handles the missing parts of array recursion.

@mvertes This contains `unsafe` when trying to reference the interface, but I see no other way to make it work.


nrwiersma:
Closed in favour of #734

PR #704, opened 2020-06-20
feature: add Sizeof and Alignof to unsafe

nrwiersma:
This adds `Sizeof` and `Alignof` to `unsafe`. Without access to the `node`, I dont see a way to add `Offsetof`, so it is left out for now.

The hook logic was also moved out of the interpreter, to keep it a little cleaner. This should also make any further extensions easier to fit in without changes to the interpreter code.

### DIFF
--- a/_test/unsafe5.go
+++ b/_test/unsafe5.go
@@ -1,0 +1,22 @@
+package main
+
+import (
+	"fmt"
+	"unsafe"
+)
+
+type S struct {
+	X int
+	Y int
+	Z int
+}
+
+func main() {
+	size := unsafe.Sizeof(S{})
+	align := unsafe.Alignof(S{})
+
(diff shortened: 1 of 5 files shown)


For-ACGN:
```go
package main

import (
	"fmt"
	"reflect"
	"unsafe"
)

type foo struct {
	A string
	B string
}

func main() {
	fmt.Println(unsafe.Offsetof(foo{}.B)) // 16

	structField, ok := reflect.TypeOf(foo{}).FieldByName("B")
	fmt.Println(structField.Offset, ok) // 16 ok
}
```
This code can get structure field offset, but i don't know how to add it to the yaegi interpreter.
I can only add a simple api but with two parameter,   offset(s interface{}, field string)  to "stdlib/unsafe".

PR #709, opened 2020-06-22
use built in golang tooling to resolve imports

### LINKED ISSUE #422: import statements do not work w/ vendored packages, go modules.
golangs package resolution is fairly complicated and the simplistic code in
```
func pkgDir(gopath, root, path string) (string, string, error) {
```
doesn't work well and misses edge cases. consider instead:

```
func pkgDir(ctx build.Context, rPath, pkg string) (string, error) {
	pkg, err := ctx.Import(path, ".", build.FindOnly)
        if err != nil {
           return "", err
        }
        return pkg.Dir, err
}
```

note: this doesn't work out of the box. I havent figured out what the other assumptions yaegi is making when resolving imports, but it at least resolves the directory of the package correctly in all cases. (including modules)

james-lawrence:
fixes #422 but using the golang toolings builtin path resolution instead of the logic in pkgDir.
could use a once over in a old style gopath environment not 100% sure on using "." for the srcDir within the yaegi context.

when yaegi interpreter is running in a go module environment pkgDir cannot locate packages modules because they are not located anywhere in the gopath.


mvertes:
I do not understand what problem the change you propose is fixing. Could you provide a test which fails without the change, and succeeds with ?

Regarding the GOPATH, it must be specified at init of interpreter to be active, otherwise only vendoring is used to search source imports. Refer to [cmd/yaegi/yaegi.go:115](https://github.com/containous/yaegi/blob/master/cmd/yaegi/yaegi.go#L115) for an example.

james-lawrence:
tl;dr; yaegi is reinventing the wheel poorly within the pkgdir, just the stdlib to do the work. its correct and more broadly used, and maintained automatically by golang devs.

tests I just didn't update, will be working on them this weekend. sorry. shouldn't have opened the PR until I finished validating I just didn't want to forget about doing it after i solved the issue for my project.

the issue is if yaegi is running inside of a project using gomodules it doesn't matter if GOPATH is set because all the project code and its vendored code have nothing to do with the GOPATH.

i.e.)
GOPATH=${HOME}/go
PROJECT_DIR=${HOME}/development/foo

all the code for project foo is going to be located directly under foo, and foo/vendored. since its not on the gopath it must be using go modules. yaegi's pkgDir has the hard coded assumption that its running inside a single gopath. it also just so happens that yaegi doesn't even support GOPATH properly, as GOPATH can have multiple directories.

james-lawrence:
probably related: #717, due to the fact yaegi's pkg resolution is broken for a number of valid golang environments.

interp/src.go
```diff
@@ -26,11 +27,14 @@ func (interp *Interpreter) importSrc(rPath, path string) (string, error) {
...
 		dir = filepath.Join(filepath.Dir(interp.Name), rPath, path)
-	} else if dir, rPath, err = pkgDir(interp.context.GOPATH, rPath, path); err != nil {
+	} else if dir, rPath, err = pkgDir(&interp.context, rPath, path); err != nil {
 		return "", err
 	}
 
 	if interp.rdir[path] {
+		// BUG: importing source can trigger panics, which leaves
+		// interp.rdir in a dirty state causing this error to be incorrect
+		// if the package failed to be imported.
```
mpl, line 37: I think what you're saying here is what I've just noticed by testing this PR: with a simple main program, that depends on `github.com/davecgh/go-spew/spew`, a valid go.mod, and GO111MODULE=on, I get a panic. Without this PR, in the same conditions, I would get the error :
`./main.go:4:2: import "github.com/davecgh/go-spew/spew" error: unable to find source related to: "github.com/davecgh/go-spew/spew"`

So I don't think this is an acceptable regression. If this PR does not fully add support for modules (which seems to be the case, given the trivial example I've tried), it should at least handle unsupported cases as gracefully as before.

mpl:
> probably related: #717 (and likely fixes)

It is indeed related, but it does not seem to fix it.
I've just tried adding the test from #717 on top of your PR, and it does not seem to pass:

```
% GO111MODULE=on CI=1 go test ./example/...
ok  	github.com/containous/yaegi/example/closure	(cached)
ok  	github.com/containous/yaegi/example/getfunc	(cached)
--- FAIL: TestPackages (0.44s)
    --- FAIL: TestPackages/eval_main_that_has_vendored_dep (0.43s)
        pkg_test.go:119: ./_pkg11/src/foo/foo.go:6:2: import "guthib.com/containous/fromage" error: unable to find source related to: "guthib.com/containous/fromage"
```

james-lawrence:
fixing modules is my motivation. i'm just pointing out all the other areas where the current code would fail under normal golang environments as well and why yaegi shouldn't be reinventing the wheel.

mpl:
> fixing modules is my motivation. i'm just pointing out all the other areas where the current code would fail under normal golang environments as well and why yaegi shouldn't be reinventing the wheel.
> 
> fixing the problem entirely with go/build package would require a larger refactor than I'm interested in working on atm as yaegi need to expose a larger api surface.
> 
> but i'll look into the panic its probably something stupid. I had to make some adjustments to my original code to get current tests passing because they violate some assumptions the go/build code makes about the work directory.

Fair enough, we'll look at the multi-valued GOPATH stuff ourselves later.
But yeah, please look into the panic as part of the scope of this PR.

interp/src.go
james-lawrence: that bug happens whenever a package can not be found, its orthogonal to the package lookup code. it exists today without these changes.

james-lawrence:
@mpl the test in #717 fails for multiple reasons.

but primarily because, like all the other current tests, it uses a relative path for gopath -- this is against golang spec.
```
GOPATH=./_pkg11/ go env
go: GOPATH entry is relative; must be absolute path: "./_pkg11/".
For more details see: 'go help gopath'
```

I was not able to reproduce any panic just the error due to the fact the test is invalid from the start.

interp/src.go
james-lawrence: releveant issue which I documented this bug over 7 months ago. https://github.com/containous/yaegi/issues/423

mvertes:
The `GoPath` option is set to `build.Default.GOPATH` in most tests under `interp/`, [here](https://github.com/containous/yaegi/blob/master/interp/interp_consistent_test.go#L107), and [here](https://github.com/containous/yaegi/blob/097a745e72d7ff5b6b542fbea76771f84f0323f9/interp/interp_file_test.go#L56), covering most tests involving files. It's also true that tests under `example/` are not updated yet and still use a relative path, and we should fix that ([here](https://github.com/containous/yaegi/blob/097a745e72d7ff5b6b542fbea76771f84f0323f9/example/closure/closure_test.go#L11), [here](https://github.com/containous/yaegi/blob/097a745e72d7ff5b6b542fbea76771f84f0323f9/example/getfunc/getfunc_test.go#L12) and [here](https://github.com/containous/yaegi/blob/097a745e72d7ff5b6b542fbea76771f84f0323f9/example/pkg/pkg_test.go#L97)).

For preciseness, we also talk about the Go Spec only referring to the [Go Programming Language Specification](https://golang.org/ref/spec), which does not cover the imports, which are specified else where, in standard library packages and / or go tools, and which can (and do) evolve faster that spec. It's always useful to point the exact reference and relevant Go versions, as it is somewhat a complex subject and moving target.

Thanks.

james-lawrence:
@mvertes so are we good on this PR? if not what needs to be done?

your last reply doesn't actually say anything beyond 'yaegi is incompatible with golang environments because it reinvented the wheel vs relying on functionality in the stdlib' which is exactly why I opened this PR.

this is not a matter of the functionality evolving, its a matter if yaegi not using APIs available in the stdlib that implement the necessary functionality and have been around since 2016.

pkgDir can be replaced with 2 lines effectively, the only reason i didn't is because yaegi now has non-standard behavior
and I'm erring on the side of caution to prevent breaking current code.

```
// pkgDir returns the absolute path in filesystem for a package given its name and
// the root of the subtree dependencies.
func pkgDir(ctx *build.Context, root, path string) (pdir string, proot string, err error) {
	pkg, err := ctx.Import(path, ".", build.FindOnly)
	return pkg.Dir, pkg.Root, err
}
```

mvertes:
I'm not ready to incorporate this PR right now. Your tone and the arguments you provide are not helpful to us and do not lead to productive discussions. We are of course aware of `go/build`. We are of course aware of duplicate design/code/efforts when they occur, and we also know about the existence of Go standard and specs. As a start, you maybe could try to understand why we chose to do otherwise than you feel right. Hint, it is not for the pleasure of re-inventing the wheel.

james-lawrence:
All yours then. I'll maintain the patch separately for my usecases. until yaegi either diverges completely or figures out what they are doing w/ respect to actually be compatible with the go ecosystem.

PR #717, opened 2020-06-25
interp: support another vendoring case

mpl:
Namely, when there's a main at the root of the project/module, and it
has direct 3rd party deps.


james-lawrence:
see #709 probabily fixes your issue as well.

PR #724, opened 2020-06-29
chore: remove solaris build due to failure

mvertes:
chore: remove solaris build due to failure

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -11,7 +11,7 @@ builds:
 #      - windows
       - freebsd
       - openbsd
-      - solaris
+#      - solaris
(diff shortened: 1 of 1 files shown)


mvertes:
The following fails with go1.14.4:
```console
$ GOOS=solaris GOARCH=amd64 go build ./...
syscall.Syscall6·f: relocation target syscall.Syscall6 not defined
```

mvertes:
It seems related to https://github.com/golang/go/issues/24357. I will change the fix and exclude some symbols known to fail in goexports.

mvertes:
Closing in favor of #726.

PR #729, opened 2020-06-30
fix: untyped constant converson to default type

### LINKED ISSUE #727: using int constant

```go
package main

import "compress/flate"

func f1(i int) { println("i:", i) }

func main() {
    i := flate.BestSpeed
    f(i)
}
```

Expected result:
```console
$ go run ./sample.go
i: 1
```

Got:
```console
$ yaegi ./sample.go
./sample.go:8:7: panic
reflect.Set: value of type constant.int64Val is not assignable to type int
goroutine 1 [running]:
runtime/debug.Stack(0x1, 0xc0002ea600, 0x40)
        /usr/lib/go/src/runtime/debug/stack.go:24 +0x9d
github.com/containous/yaegi/interp.(*Interpreter).Eval.func1(0xc000227de0)
        /home/marc/go/src/github.com/containous/yaegi/interp/interp.go:346 +0xbc
panic(0xd3ba80, 0xc000115550)
        /usr/lib/go/src/runtime/panic.go:969 +0x166
github.com/containous/yaegi/interp.runCfg.func1(0xc00036c0a0, 0xc000375f00)
        /home/marc/go/src/github.com/containous/yaegi/interp/run.go:115 +0x20c
panic(0xd3ba80, 0xc000115550)
        /usr/lib/go/src/runtime/panic.go:969 +0x166
reflect.Value.assignTo(0xdaaec0, 0xc000368320, 0x186, 0xe79a00, 0xb, 0xd3ad40, 0x0, 0xd1e660, 0x1, 0x6)
        /usr/lib/go/src/reflect/value.go:2403 +0x426
reflect.Value.Set(0xd3ad40, 0xc000368328, 0x182, 0xdaaec0, 0xc000368320, 0x186)
        /usr/lib/go/src/reflect/value.go:1532 +0xbd
github.com/containous/yaegi/interp.call.func7(0xc00036c0a0, 0xc00036b730)
        /home/marc/go/src/github.com/containous/yaegi/interp/run.go:1001 +0x98f
github.com/containous/yaegi/interp.runCfg(0xc000375f00, 0xc00036c0a0)
        /home/marc/go/src/github.com/containous/yaegi/interp/run.go:121 +0x6d
github.com/containous/yaegi/interp.(*Interpreter).run(0xc0001381a0, 0xc000375700, 0xc00036c000)
        /home/marc/go/src/github.com/containous/yaegi/interp/run.go:99 +0x239
github.com/containous/yaegi/interp.(*Interpreter).Eval(0xc0001381a0, 0xc0002dd280, 0x79, 0x0, 0x0, 0x0, 0x0, 0x0)
        /home/marc/go/src/github.com/containous/yaegi/interp/interp.go:422 +0x52b
main.main()
        /home/marc/go/src/github.com/containous/yaegi/cmd/yaegi/yaegi.go:159 +0x65d

```

Interestingly, calling `f1(flate.BestSpeed)` instead of `f1(i)` works as expected.

mvertes:
In definition assign expression, the source type is propagated to
the assigned value. If the source is an untyped constant, the
destination type must be set to the default type of the constant
definition. A fixType function is provided to perform this.

In addition, the type conversion and check of constants is
refactored for simplifications.

Fixes #727.


interp/run.go
```diff
@@ -2824,86 +2735,62 @@ func convertConstantValueTo(n *node, typ reflect.Type) {
...
-			}
-			v = reflect.ValueOf(i)
+		if l > bitlen[kind] {
+			panic(fmt.Sprintf("constant %s overflows uint%d", c.ExactString(), bitlen[kind]))
 		}
+		v = reflect.ValueOf(i).Convert(typ)
 	case reflect.Float32:
 		f, _ := constant.Float32Val(c)
-		v = reflect.ValueOf(f)
+		v = reflect.ValueOf(f).Convert(typ)
```
nrwiersma, line 2761: Do we need the `Convert` here? Also float64 and complex128?

interp/run.go
```diff
@@ -2824,86 +2735,62 @@ func convertConstantValueTo(n *node, typ reflect.Type) {
...
-				panic(fmt.Sprintf("constant %s overflows int32", c.ExactString()))
-			}
-			v = reflect.ValueOf(int32(i))
-		case reflect.Int64:
-			if l > 64 {
-				panic(fmt.Sprintf("constant %s overflows int64", c.ExactString()))
-			}
-			v = reflect.ValueOf(i)
+		if l > bitlen[kind] {
+			panic(fmt.Sprintf("constant %s overflows int%d", c.ExactString(), bitlen[kind]))
```
nrwiersma, line 2749: the error will be wrong in the case of `int` which would show up as `int64`. Not an actual issue, but perhaps confusing?

interp/run.go
mvertes: I think so. We may have an imported typed constant definition or expression (with type propagation), with an aliased type on `float32`.

interp/run.go
mvertes: A small cost to pay for the code simplification it allows.

nrwiersma: agreed.

interp/run.go
```diff
@@ -2824,86 +2735,62 @@ func convertConstantValueTo(n *node, typ reflect.Type) {
...
 		i, _ := constant.Float64Val(constant.Imag(c))
-		v = reflect.ValueOf(complex(r, i))
+		v = reflect.ValueOf(complex(r, i)).Convert(typ)
+	default:
+		// Type kind is from internal constant representation. Only use default types here.
+		switch c.Kind() {
+		case constant.Bool:
+			v = reflect.ValueOf(constant.BoolVal(c))
+		case constant.String:
+			v = reflect.ValueOf(constant.StringVal(c))
```
nrwiersma, line 2779: These 2 cases are impossible at this point. We only have `constant.Value` values for numeric kinds. Should we still support it?

mvertes: I think that keeping it causes no harm, and I wonder also if we could come out with a constant boolean expression involving large numbers where this could be useful (not tested yet).

nrwiersma: I dont think these cases are handled in the ops, but that that another PR.

PR #741, opened 2020-07-03
feature: sort vars by dependencies

nrwiersma:
This implements dependency resolution ordering for vars at the global level. In the case of a var with a call expression, the order of execution is important. Perhaps this should be added to `cfg` and be at all levels?

@mvertes I would like your opinion on this before making it a full PR. I also think we can clean it up a bit, but that will be done before opening the PR.

### DIFF
--- a/_test/var12.go
+++ b/_test/var12.go
@@ -1,0 +1,24 @@
+package main
+
+var A = concat("hello", B)
+
+var B = concat(" ", C, "!")
+
+var C = D
+
+var D = "world"
+
+func concat(a ...string) string {
+	var s string
+	for _, ss := range a {
+		s += ss
+	}
+	return s
+}
+
+func main() {
+	println(A)
+}
+
+// Output:
+// hello world!
--- a/interp/ast.go
+++ b/interp/ast.go
@@ -666,6 +666,15 @@ func (interp *Interpreter) ast(src, name string) (string, *node, error) {
 			case token.VAR:
 				kind = varDecl
 			}
+
+			if kind == varDecl {
+				// In order to group all vars and consts under a single declaration
+				// find the correct anc.
+				if n := findNodeKind(anc.node.child, kind); n != nil {
+					st.push(n, nod)
+					break
+				}
+			}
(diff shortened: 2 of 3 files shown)


mvertes:
A few remarks:
1. The test provided here deserves a dedicated issue, independently of this PR.
2. A solution must also take into account the fact that the separate declarations can be spread out in different files, on which you have no control of parsing order. It seems to be not taken into account here. Therefore modidying the AST is not a good idea here.
3. Note that just replacing `var` by `const` in the tests (keeping the _wrong_ order) avoids the constant definition loop error. The result is still not correct at the end (and is rejected by `go run` as it is a function call is not considered as a constant expression). 

I think a variant of the constant definition algorithm applied to gloval variables will give you the right sequence order (may be with an additional field in `symbol` to track the dependency. This order should be applied to the children of `fileStmt` node, and this is to be done in post-processing in `cfg.go` for a node of `fileStmt` kind, when setting the `node.next` value (which node should be executed next), without having to change the AST.

mvertes:
In fact, the rewiring of `next` nodes should be done after the `CFG` pass, in `eval.go`  and  in `src.go`, and not in `cfg.go` itself, where the AST is still a partial one.

nrwiersma:
At the end of the day, I also came to the same conclusion. Mainly as I dont know what is a var ident and what is not. The change to the AST is only to group all vars under a single varDecl for the same ancestor. This is mainly to make processing this easier, as even in post order CFG, keeping track of the correct order might be interesting. I will try it without and make and issue for the example.

At this point I mainly wanted to validate the validity of the idea of this, hence the no issue and dirty test code.

nrwiersma:
That is an interesting thought, doing it after all CFG. The issue I see there is having to parse the entire node tree over again. I was trying to do it inline to eliminate this need. I also noted that in a multi file import, if the symbol for the var is not found, the file is revisited, meaning cross file declarations should not be an issue, but I have not tested it yet.

nrwiersma:
Ok, after reading the spec, I understand the procedure needed for package level vars. I will relook at this.

PR #753, opened 2020-07-07
extract: new package to extract symbols from a dependency

mpl:
cmd/goexports is now based on it.

Known limitations for now:
- dependency given as argument needs to be in GOPATH
- vendoring not supported
- requires GO111MODULE=off


mpl:
The CI is failing because the new goexports generates a different stdlib/go1_14_math.go, which I didn't include in the PR, since I don't understand why yet. So the git indexes differ.

PR #771, opened 2020-07-10
fix: type assertion expression was not forwarding type

### LINKED ISSUE #770: inducing type from type assert at parse
The following program `sample.go` triggers an unexpected:

```go
import "reflect"

type I interface {
        Foo() int
}

type T struct {
        Name string
}

func (t T) Foo() int { return 0 }

func f(v reflect.Value) int {
        i := v.Interface().(I)
        return i.Foo()
}

func main() {
        println("hello")
}
```

Expected result:
```console
$ go run ./sample.go
hello
```

Got:
```console
$ yaegi ./sample.go
./sample.go:17:9: undefined field or method: Foo
```

mvertes:
The detection of special cases of interpreter functions and interfaces
is more precise. It allows at least to complete parsing of
code where type is derived from a type assertion expression.

Fixes #770.


interp/cfg.go
```diff
@@ -1724,17 +1724,18 @@ func (interp *Interpreter) cfg(root *node, pkgID string) ([]*node, error) {
 		case typeAssertExpr:
 			if len(n.child) > 1 {
 				wireChild(n)
-				if n.child[1].typ == nil {
-					if n.child[1].typ, err = nodeType(interp, sc, n.child[1]); err != nil {
+				c1 := n.child[1]
+				if c1.typ == nil {
+					if n.child[1].typ, err = nodeType(interp, sc, c1); err != nil {
```
mpl, line 1729: why not:
```suggestion
					if c1.typ, err = nodeType(interp, sc, c1); err != nil {
```
?

PR #773, opened 2020-07-10
interp: support yet another vendoring case

### LINKED ISSUE #758: Vendoring management
Hi,

When you need vendors in a subfolder inside your code, yaegi doesn't find them. You have to copy the complete vendor folder to the subfolder (a symlink doesn't work).

Pierre

mpl:
Namely, when the vendor dir is a sibling (or an uncle) relative to the
current pkg

Fixes #758

### DIFF
--- a/interp/src.go
+++ b/interp/src.go
@@ -1,9 +1,11 @@
 package interp
 
 import (
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"os"
+	"path"
 	"path/filepath"
 	"strings"
 )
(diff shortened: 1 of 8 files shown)


interp/src.go
```diff
@@ -188,13 +190,87 @@ func pkgDir(goPath string, root, path string) (string, string, error) {
...
+
+	const vendor = "vendor"
+
+	// TODO(mpl): maybe it works for the special case main, but can't be bothered for now.
+	parent, final := filepath.Split(rootPath)
+	if root != mainID && final != vendor {
+		// look for the closest vendor in one of our direct ancestors, as it takes priority.
+		prefix := path.Clean(strings.TrimSuffix(rootPath, root))
+		var vendored string
+		for {
```
mvertes, line 216: Here, you do not need to open the directories at each level and read/parse potentially all of their content. You could just construct the next path (up to `/`) and probe if the `path/vendor` is a valid directory using `os.Lstat` and `IsDir` method. That would simplify a lot (no defer, no closure, ...) and be always faster.

I do not see the need of `depthCounter` as well, as the number of paths up to `/` is always finite, and we are not walking through symlinks etc.

interp/src_test.go
```diff
@@ -222,7 +222,9 @@ func Test_previousRoot(t *testing.T) {
 		t.Run(test.desc, func(t *testing.T) {
 			t.Parallel()
 
-			p := previousRoot(test.root)
+			// TODO(mpl): add tests for new algo.
+			// it's not too bad to not have them for now, since the tests in example run this code path too.
+			p, _ := previousRoot("vendor", test.root)
```
mvertes, line 227: I do not get why you pass `"vendor"` here as it is defined  and used as a const in `previousRoot`.

You have also changed the `previousRoot` signature to return an error, but you are not testing or using it.

interp/src.go
mpl: > Here, you do not need to open the directories at each level and read/parse potentially all of their content. You could just construct the next path (up to `/`) and probe if the `path/vendor` is a valid directory using `os.Lstat` and `IsDir` method. That would simplify a lot (no defer, no closure, ...) and be always faster.

Agreed.

> I do not see the need of `depthCounter` as well, as the number of paths up to `/` is always finite, and we are not walking through symlinks etc.

Probably, yes. As this needed to be finished today (I won't be back until thursday, and they needed this fixed very soon) I opted for being overly safe just in case, and not spend too much time thinking about it.

interp/src_test.go
mpl: Passing "vendor" as argument should guarantee that the new code is _not_ used for these tests. (Ah, if you mean that we could use a const instead of an hardcoded literal, then yes, sure. The const was not in my head, I added it because the linter forced me to). Again, I was pressed by time, and just wanted to be sure to keep these tests passing. Hence the TODO for later.

interp/src.go
mpl: All done

interp/src_test.go
mpl: done

PR #786, opened 2020-07-21
chore: add homebrew-tap recipe

kamilsk:
Publish homebrew-tap recipe to https://github.com/traefik/homebrew-tap

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -45,3 +45,24 @@ archives:
         format: zip
     files:
       - LICENSE
+
+brews:
+  - name: yaegi
+    caveats:
+    commit_author:
+      name:  <TODO>
+      email: <TODO>
+    description: |
+      Yaegi is Another Elegant Go Interpreter.
+      It powers executable Go scripts and plugins, in embedded interpreters
+      or interactive shells, on top of the Go runtime.
+    folder: Formula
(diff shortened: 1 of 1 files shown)


.goreleaser.yml
```diff
@@ -70,3 +70,24 @@ archives:
         format: zip
     files:
       - LICENSE
+
+brews:
+  - name: yaegi
+    caveats:
+    commit_author:
+      name:  <TODO>
+      email: <TODO>
```
kamilsk, line 79: do we need to configure this or use defaults?

.goreleaser.yml
```diff
@@ -70,3 +70,24 @@ archives:
...
+    description: |
+      Yaegi is Another Elegant Go Interpreter.
+      It powers executable Go scripts and plugins, in embedded interpreters
+      or interactive shells, on top of the Go runtime.
+    folder: Formula
+    homepage: https://github.com/containous/yaegi
+    github:
+      owner: <TODO>
+      name:  <TODO>
```
kamilsk, line 88: and what repo I have to use to push Formula on successful builds?

mvertes:
As we plan to migrate/rename some repos, we will apply this later, to avoid to do twice the same thing.

ldez (approved):
LGTM

PR #795, opened 2020-07-30
interp: change API of Eval

### LINKED ISSUE #731: Redeclaring imports
I used to load multiple scripts using **Eval** method in version 0.8.5. They had same package name and shared some same imports. Now I'm getting 'xxx redeclared in this block' (xxx being the imported package name).
Not sure why is this invalid now as this feature allowed me to split sources across multiple files.

### LINKED ISSUE #778: redeclared struct

yaegi version: v0.8.9 -  v0.8.12 

```go
package yaegi

import (
	"testing"
)

const src = `package main

import (
)

type GetHistoryLocRsp struct { 
    DailyItems []DailyItem  
}

type DailyItem struct { 
    Timestamp int64    
}
`

func TestM(tt *testing.T) {
	_, err := i.Eval(src)
	if err != nil {
		panic(err)
	}
}
```

Expected result:
```console
the code will run without panic

```

Got:
```console
panic：

panic: 10:6: DailyItem redeclared in this block [recovered]
	panic: 10:6: DailyItem redeclared in this block

goroutine 270 [running]:
testing.tRunner.func1.1(0x1b55740, 0xc00017bae0)
	/home/yuanyangen/.install/go/src/testing/testing.go:940 +0x4dc
testing.tRunner.func1(0xc0003cc900)
	/home/yuanyangen/.install/go/src/testing/testing.go:943 +0x317
panic(0x1b55740, 0xc00017bae0)
	/home/yuanyangen/.install/go/src/runtime/panic.go:975 +0x41b
code.byted.org/hotsoon/func_core/executor/yaegi.TestM(0xc0003cc900)
	/home/yuanyangen/HomeData/go/func/func_core/executor/yaegi/yaegi_raw_test.go:26 +0x10e
testing.tRunner(0xc0003cc900, 0x1c72670)
	/home/yuanyangen/.install/go/src/testing/testing.go:991 +0x18a
created by testing.(*T).Run
	/home/yuanyangen/.install/go/src/testing/testing.go:1042 +0x660
```

mpl:
The recent changes that added some redeclaration checks implicitly added more
strictness related to namespaces and scopes which, among other things, broke
some uses that "accidentally" used to work.

For example, given

```
const script1 = `
	import "fmt"

	// more code
`
const script2 = `
	import "fmt"

	// some other code
`
```

If one Evals script1, then script2, with the same interpreter, without
specifying any scope, as the two fragments would be considered part of the same
(.go file) scope by default, a redeclaration error would be triggered because
`import "fmt"` is seen twice.

A work-around would have been to specify (a different) `i.Name` before each Eval
call, so that each script is considered as coming from a different .go file, and
hence are respectively in different scopes with respect to imports.

That lead us to realize we had to make specifying things such as file-scope, and
"incremental mode" (aka REPL), more obvious in the context of an Eval call.
Hence the new signature for Eval:

```
func (interp *Interpreter) Eval(src, name string, incremental bool) (res reflect.Value, err error)
```

And to avoid confusion, the `Name` field of Interpreter is now non-exported,
since it would be redundant with the `name` argument of Eval.

As a convenience, the old Eval still exists, as:

```
func (interp *Interpreter) EvalInc(src string) (res reflect.Value, err error)
```

Finally, some related bugfixes, documention changes, and some refactoring have
been included. Notably, there is no "empty scope" anymore, i.e. `name` defaults
to "_.go" when it is not specified.

Fixes #731
Fixes #778

### DIFF
--- a/_test/eval0.go
+++ b/_test/eval0.go
@@ -9,13 +9,13 @@ import (
 func main() {
 	log.SetFlags(log.Lshortfile)
 	i := interp.New(interp.Options{})
-	if _, err := i.Eval(`func f() (int, int) { return 1, 2 }`); err != nil {
+	if _, err := i.EvalInc(`func f() (int, int) { return 1, 2 }`); err != nil {
 		log.Fatal(err)
 	}
-	if _, err := i.Eval(`a, b := f()`); err != nil {
+	if _, err := i.EvalInc(`a, b := f()`); err != nil {
 		log.Fatal(err)
 	}
-	if _, err := i.Eval(`println(a, b)`); err != nil {
+	if _, err := i.EvalInc(`println(a, b)`); err != nil {
 		log.Fatal(err)
 	}
 }
--- a/_test/inception.go
+++ b/_test/inception.go
@@ -10,13 +10,13 @@ func main() {
 	log.SetFlags(log.Lshortfile)
 	i := interp.New(interp.Options{})
 	i.Use(interp.Symbols)
-	if _, err := i.Eval(`import "github.com/containous/yaegi/interp"`); err != nil {
+	if _, err := i.EvalInc(`import "github.com/containous/yaegi/interp"`); err != nil {
 		log.Fatal(err)
 	}
-	if _, err := i.Eval(`i := interp.New(interp.Options{})`); err != nil {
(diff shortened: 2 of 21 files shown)


mpl:
Closing in favor of #802 because I messed up with branches.

PR #802, opened 2020-08-05
interp: change API of Eval

### LINKED ISSUE #731: Redeclaring imports

### LINKED ISSUE #778: redeclared struct

### LINKED ISSUE #789: Maybe infinite loop while looking for missing vendored dependencies
TODO(mpl): to be confirmed/described

Hint: when a plugin looks for its deps in the vendor dir, but they are absent.

### LINKED ISSUE #798: interp: another undetected redeclaration
```
mbp:foo mpl$ tree
.
├── foo.go
└── vendor
    └── guthib.com
        ├── bar
        │   └── quux.go
        └── baz
            └── quux.go
```

```go
$ cat vendor/guthib.com/bar/quux.go
package quux

func Quux() string {
	return "bar"
}
```

```go
$ cat vendor/guthib.com/baz/quux.go
package quux

func Quux() string {
	return "baz"
}
```

```go
$ cat foo.go
package main

import (
	"guthib.com/bar"
	"guthib.com/baz"
)

func main() {
	println("Hello", bar.Quux())
	println("Hello", baz.Quux())
}
```

Expected result:
```console
$ go run ./foo.go
# command-line-arguments
./foo.go:5:2: quux redeclared as imported package name
	previous declaration at ./foo.go:4:2
./foo.go:5:2: imported and not used: "foo/vendor/guthib.com/baz" as quux
./foo.go:9:19: undefined: bar
./foo.go:10:19: undefined: baz
```

Got:
```console
$ yaegi ./foo.go
Hello bar
Hello baz
```

mpl:
The recent changes that added some redeclaration checks implicitly added more
strictness related to namespaces and scopes which, among other things, broke
some uses that "accidentally" used to work.

For example, given

```
const script1 = `
	import "fmt"

	// more code
`
const script2 = `
	import "fmt"

	// some other code
`
```

If one Evals script1, then script2, with the same interpreter, without
specifying any scope, as the two fragments would be considered part of the same
(.go file) scope by default, a redeclaration error would be triggered because
`import "fmt"` is seen twice.

A work-around would have been to specify (a different) `i.Name` before each Eval
call, so that each script is considered as coming from a different .go file, and
hence are respectively in different scopes with respect to imports.

That lead us to realize we had to make specifying things such as file-scope, and
"incremental mode" (aka REPL), more obvious in the context of an Eval call.

In addition, we want to lay down the foundations for Yaegi being able to behave
more like the go tool wrt to various inputs, i.e. it should be able to take a
package directory, or an import path, as input, instead of just a .go file.

Hence the introduction of a new kind of Eval method (whose signature is not fixed yet):

```
func (interp *Interpreter) EvalPath(path string) (res reflect.Value, err error)
```

It partially solves the problem described above because:
1) the path given to EvalPath can be used as the file-scope hint mentionned
above, for now (even though the related implementation details might change).
2) Eval always runs in incremental mode, whereas EvalPath always runs in
non-incremental mode, hence clarifying the sitation in that respect.

And to avoid confusion, the `Name` field of Interpreter is now non-exported,
since it is somewhat redundant with the `path` argument of EvalPath.

Note that #731 is not fully fixed (and might never be), as a requirement of the
proposed solution is to move the input bits of code into respective files
(instead of leaving them as strings).

Finally, some related bugfixes, documention changes, and some refactoring have
been included. Notably, there is no "empty scope" anymore, i.e. `name` defaults
to "_.go" when it is not specified.

Updates #731
Fixes #778
Fixes #798 
Fixes #789


mvertes:
After rebasing, and following discussions with @mpl, I also reverted the introduction of `EvalInc`, keeping `Eval` instead. I have also introduced `EvalName` and `EvalPath` public methods and `eval` private method. I suppressed the `inREPL` field in `Interpreter` struct.

mvertes (approved):
LGTM

interp/interp.go
```diff
@@ -330,26 +329,38 @@ func (interp *Interpreter) main() *node {
...
 
-// Eval evaluates Go code represented as a string. It returns a map on
-// current interpreted package exported symbols.
-// Name is the file path to the Go source file containing src. It defaults to the
-// special value "_.go".
-// Incremental is whether the given source should be considered an integral
-// valid Go source file, or only a fragment.
-func (interp *Interpreter) Eval(src, name string, incremental bool) (res reflect.Value, err error) {
+// EvalName evaluates Go code represented as a string. Name is the import path of code
+// and can be left to "" for main. EvalName returns the last result computed by the
```
mpl, line 339: "can" or "should" ?

interp/interp.go
```diff
@@ -330,26 +329,38 @@ func (interp *Interpreter) main() *node {
...
+// interpreter, and a non nil error in case of failure.
+func (interp *Interpreter) EvalName(src, name string) (res reflect.Value, err error) {
+	return interp.eval(src, name, true)
+}
+
+// EvalPath evaluates Go file by path. EvalPath returns the last result
```
mpl, line 345:
```suggestion
// EvalPath evaluates the Go code located at path. EvalPath returns the last result
```

interp/interp.go
```diff
@@ -330,26 +329,38 @@ func (interp *Interpreter) main() *node {
...
-// Name is the file path to the Go source file containing src. It defaults to the
-// special value "_.go".
-// Incremental is whether the given source should be considered an integral
-// valid Go source file, or only a fragment.
-func (interp *Interpreter) Eval(src, name string, incremental bool) (res reflect.Value, err error) {
+// EvalName evaluates Go code represented as a string. Name is the import path of code
+// and can be left to "" for main. EvalName returns the last result computed by the
+// interpreter, and a non nil error in case of failure.
+func (interp *Interpreter) EvalName(src, name string) (res reflect.Value, err error) {
+	return interp.eval(src, name, true)
```
mpl, line 342: Why is EvalName always in incremental mode?
And if it is, I think it's weird that some of the Eval* methods are in incremental mode, whereas some others are not, and that we haven't even documented it for the user.

interp/src.go
```diff
@@ -88,26 +100,27 @@ func (interp *Interpreter) importSrc(rPath, path string) (string, error) {
...
 		if err != nil {
 			return "", err
 		}
 		revisit[subRPath] = append(revisit[subRPath], list...)
 	}
 
 	// Revisit incomplete nodes where GTA could not complete.
-	for pkg, nodes := range revisit {
-		if err = interp.gtaRetry(nodes, pkg, path); err != nil {
+	//	for pkg, nodes := range revisit {
```
mpl, line 113: TODO(mpl): remove

PR #817, opened 2020-08-14
interp: take into account embedded property of struct field

### LINKED ISSUE #781: can not the same json marshal result as same as golang
```go
package yaegi

import (
	yaegi_stdlib "github.com/containous/yaegi/stdlib"
	"testing"
)

const src = `package main

import (
	"encoding/json"
	"fmt"
)

type A struct { 
   	InnerA
}

type InnerA struct { 
    Timestamp int64    
}

func main() {
	a := &A{}
	b, _ := json.Marshal(a)
	fmt.Println(string(b))
}
`

func TestM(tt *testing.T) {
	i.Use(yaegi_stdlib.Symbols)
	_, err := i.Eval(src)
	if err != nil {
		panic(err)
	}
}
```

Expected result:
```console
the json marshal result in `src` should be:

{"Timestamp":0}

```

Got:
```console
{"InnerA":{"Timestamp":0}}
```

mpl:
The trick is that in reflect, the field is called Anonymous, which is
actually a different notion in the Go spec/vocable.

n.b. only works for non-recursive types for now.

Fixes #781


mvertes (approved):
LGTM

PR #818, opened 2020-08-17
feat: add slice expression type checking

nrwiersma:
This adds type checking to SliceExpr as well as handling any required constant type conversion.

It should be noted that the check that `high` and `max` are not nil in a 3-index slice has been left out as this check is handled before type checking. Tests have been included for these cases.


mvertes (approved):
LGTM

PR #821, opened 2020-08-19
feat: add type assertion expression type checking

nrwiersma:
This adds type checking to TypeAssertExpr. In order to allow for this, method types now have a receiver type in both reflect and native cases.


mvertes (approved):
LGTM

PR #823, opened 2020-08-20
feature: improve handling of interrupt signal in REPL

mvertes:
The input scanning is now performed in a sub goroutine and
the interrupt is listened in another goroutine, either to cancel Eval
or to cancel the current line scan.
Entering a '\n' after a 'Ctrl-C` to get the prompt is
not necessary anymore.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -615,11 +615,13 @@ func (interp *Interpreter) REPL(in io.Reader, out io.Writer) {
 	}()
 
 	for {
+		var line string
+
 		select {
 		case <-end:
 			cancel()
 			return
-		case line := <-lines:
+		case line = <-lines:
 			src += line + "\n"
 		}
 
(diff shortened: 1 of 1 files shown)


interp/interp.go
```diff
@@ -452,8 +452,8 @@ func (interp *Interpreter) EvalWithContext(ctx context.Context, src string) (ref
 		interp.stop()
 		return reflect.Value{}, ctx.Err()
```
mpl, line 453: How about
```suggestion
		v, err = reflect.Value{}, ctx.Err()
```
instead? That would make L456 the unique return point, which would be more readable and elegant imho.

mpl (changes requested):

interp/interp.go
mvertes: This change causes a race condition, due to the concurrency with line 447.

mpl (approved):

PR #824, opened 2020-08-21
REPL: retry with full wrapping for anonymous func calls

### LINKED ISSUE #721: In-place func calls parsing in REPL or -e string
In-place func calls in REPL or `-e string` fail to be parsed when no extra parentheses around `func(){}` is given.

Expected result:
```bash
$ yaegi -i -e 'func(){println(0)}()'
0
> (func(){println(1)})()
1
: 0xc00006c2c0
> func(){println(2)}()
2
: 0xwhateverhex
> █
```

Got:
```bash
$ yaegi -i -e 'func(){println(0)}()'
> (func(){println(1)})()
1
: 0xc00006c2c0
> func(){println(2)}()
█
```

mpl:
In interactive mode, a line starting with the "func" keyword is usually
"wrapped", by prepending to it a "package main" statement, to make it
a valid piece of Go source code.

However, when the line is actually an anonymous function call, such as:

```
func() { println(3) }()
```

then this wrapping is not enough, as this is not valid Go in the global
context. Therefore, this kind of of expression must also be wrapped
inside a main func (as is the default case for most REPL inputs).

Since the detection and handling of such a case turned out to be quite
unelegant, this PR instead introduces a retrying phase when a parsing
error occurs for a particular class of cases. That is to say, when a
"func expression" wrapped in a main package fails to be parsed, it is
then wrapped in a main func before parsing is retried.

Fixes #721

### DIFF
--- a/interp/ast.go
+++ b/interp/ast.go
@@ -348,18 +348,32 @@ func (interp *Interpreter) firstToken(src string) token.Token {
 // The given name is used to set the filename of the relevant source file in the
 // interpreter's FileSet.
 func (interp *Interpreter) ast(src, name string, inc bool) (string, *node, error) {
-	var inFunc bool
+	var inFunc, shouldRetry bool
+	var initialError error
+	initialSrc := src
 	var mode parser.Mode
 
+LREPL:
 	// Allow incremental parsing of declarations or statements, by inserting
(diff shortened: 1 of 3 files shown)


mvertes (approved):
LGTM

mvertes (approved):
LGTM

PR #826, opened 2020-08-24
fix: check array size symbol kind

### LINKED ISSUE #825: non-constant array bound
The following code  should trigger  "non-constant array bound xte"
```go
xte:=10
ute:=[xte]byte{1,2}
fmt.Println(ute)
```

however it compiles and displays the result in yaegi.
However leaving it as it is, could be used for creating maps with dynamic length arrays as keys.

nrwiersma:
When determining the size of an array and a symbol is found, the symbol must be a const for the type to be valid. 

While it makes sense for this check to be done in type checking, the type can be determined by GTA which would then fail. For now this check is done when getting the node type.

Fixes #825


mvertes (approved):
LGTM

PR #827, opened 2020-08-25
feat: add builtin type checking

nrwiersma:
This adds type checking for builtin functions. It also refactors builtin names into constants due to the number of times they are used.


mvertes (approved):
LGTM

PR #828, opened 2020-08-25
interp: enable declaration errors detection at parsing time

### LINKED ISSUE #811: interp: investigate enabling DeclarationErrors for go/parser
#### Proposal

Method ParseFile takes a mode, in which DeclarationErrors can be enabled. It seems we could simply do that instead of handling redeclaration ourselves in the post-ast stages.

#### Background

Handling redeclarations properly in Yaegi. As of now they're partially handled with custom error catching in the CFG and GTA stages.

mpl:
The Go parser is able to detect some (but not all) (re-)declaration errors,
if the DeclarationErrors flag is enabled, which was not done so far.

This PR therefore enables that flag, which allows the removal of some of
the now unneeded code that was recently added to support redeclarations.

Fixes #811

### DIFF
--- a/_test/redeclaration-global5.go
+++ b/_test/redeclaration-global5.go
@@ -12,4 +12,5 @@ func main() {
 }
 
 // Error:
-// ../_test/redeclaration-global5.go:5:1: time redeclared in this block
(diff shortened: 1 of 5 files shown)


mvertes (approved):
LGTM

PR #830, opened 2020-08-26
feat: configure stdin, stdout and stderr per interpreter

### LINKED ISSUE #752: Support Explicit Passing of Stdout/Stderr to Yaegi
#### Proposal

Add the ability to explicitly pass an `io.Writer` to the Yeagi interpreter so that terminal output from the interpreted code could be avoided. Here is some example code with my proposed change:

```go
package main

import (
	"bytes"
	"fmt"
	"os"

	"github.com/containous/yaegi/stdlib"
)

const script = `
package main

import (
	"fmt"
	"os"
)

func main() {
	fmt.Fprintln(os.Stdout, "hello from interp stdout")
	fmt.Fprintln(os.Stderr, "hello from interp stderr")
}
`

func main() {
	stdout := bytes.NewBuffer(nil)
	stderr := bytes.NewBuffer(nil)
        // if no writers are passed the interpreter should fallback
        // to writing to stdout/stderr of the calling process.
	i := interp.New(interp.Options{
		Stdout: stdout, // new functionality here
		Stderr: stderr,
	})
	i.Use(stdlib.Symbols)
	i.Eval(script)
	fmt.Fprintln(os.Stdout, "hello from main stdout")
	fmt.Fprintln(os.Stderr, "hello from main stderr")
}
```

This was originally opened in https://github.com/containous/yaegi/issues/728 by @g3rzi however they closed the issue with a workaround but I felt it was still valuable.

Thanks for your work on this great project!

#### Background

I am writing a [tool](https://github.com/kevinschoon/qviz) to do data visualization that leverages Yaegi and I would like to be able to explicitly control the stdout/stderr of evaluated programs so I can display logs in a UI.

#### Workarounds

@mpl mentions a workaround here: https://github.com/containous/yaegi/blob/4f3481b55c207b933902314a7eac6aee88ed82d6/example/pkg/pkg_test.go#L109-L135

mvertes:
The goal is to provide greater control of input, output and error
streams of the interpreter. It is now possible to specify those
as options when creating a new interpreter. The provided values
are propagated to relevant stdlib symbols (i.e fmt.Print, etc).
Care is taken to not update the global variables os.Stdout, os.Stdin
and os.Stderr, as to not interfere with the host process.

The REPL function is now simplified. The deprecated version is removed.

The tests are updated to take advantage of the simplified access
to the interpreter output and errors.

Fixes #752.

### DIFF
--- a/cmd/yaegi/run.go
+++ b/cmd/yaegi/run.go
@@ -56,14 +56,14 @@ func run(arg []string) error {
 	}
 
 	if cmd != "" {
-		i.REPL(strings.NewReader(cmd), os.Stderr)
+		_, err = i.Eval(cmd)
 	}
 
 	if len(args) == 0 {
 		if interactive || cmd == "" {
-			i.REPL(os.Stdin, os.Stdout)
+			_, err = i.REPL()
 		}
-		return nil
+		return err
 	}
 
 	// Skip first os arg to set command line as expected by interpreted main
@@ -85,9 +85,9 @@ func run(arg []string) error {
 	}
 
 	if interactive {
-		i.REPL(os.Stdin, os.Stdout)
+		_, err = i.REPL()
 	}
-	return nil
+	return err
 }
 
 func isPackageName(path string) bool {
@@ -116,7 +116,7 @@ func runFile(i *interp.Interpreter, path string) error {
 	if s := string(b); strings.HasPrefix(s, "#!") {
 		// Allow executable go scripts, Have the same behavior as in interactive mode.
 		s = strings.Replace(s, "#!", "//", 1)
-		i.REPL(strings.NewReader(s), os.Stdout)
+		_, err = i.Eval(s)
 	} else {
 		// Files not starting with "#!" are supposed to be pure Go, directly Evaled.
 		_, err := i.EvalPath(path)
@@ -127,5 +127,5 @@ func runFile(i *interp.Interpreter, path string) error {
 			}
 		}
 	}
-	return nil
+	return err
 }
--- a/cmd/yaegi/yaegi_test.go
+++ b/cmd/yaegi/yaegi_test.go
@@ -2,11 +2,13 @@ package main
 
 import (
 	"bytes"
+	"context"
 	"io/ioutil"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strconv"
+	"strings"
 	"testing"
 	"time"
 )
@@ -108,7 +110,7 @@ func TestYaegiCmdCancel(t *testing.T) {
 			continue
 		}
 
-		if errBuf.String() != "context canceled\n" {
+		if strings.TrimSuffix(errBuf.String(), "\n") != context.Canceled.Error() {
 			t.Errorf("unexpected error: %q", &errBuf)
 		}
 	}
--- a/cmd/yaegi/yaegi_test.go
+++ b/cmd/yaegi/yaegi_test.go
@@ -108,8 +108,8 @@ func TestYaegiCmdCancel(t *testing.T) {
 			continue
 		}
 
-		if outBuf.String() != "context canceled\n" {
-			t.Errorf("unexpected output: %q", &outBuf)
+		if errBuf.String() != "context canceled\n" {
+			t.Errorf("unexpected error: %q", &errBuf)
 		}
 	}
 }
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -4,12 +4,14 @@ import (
 	"bufio"
 	"context"
 	"errors"
+	"flag"
 	"fmt"
 	"go/build"
 	"go/scanner"
 	"go/token"
 	"io"
 	"io/ioutil"
+	"log"
 	"os"
 	"os/signal"
 	"reflect"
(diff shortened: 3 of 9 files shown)


cmd/yaegi/run.go
```diff
@@ -56,14 +56,14 @@ func run(arg []string) error {
 	}
 
 	if cmd != "" {
-		i.REPL(strings.NewReader(cmd), os.Stderr)
+		_, err = i.Eval(cmd)
```
mpl, line 59: Did you change that REPL call to an Eval call on purpose?

interp/interp.go
```diff
@@ -207,16 +212,32 @@ func (n *node) Walk(in func(n *node) bool, out func(n *node)) {
...
+		in = os.Stdin
+	}
+	out := options.Stdout
+	if out == nil {
+		out = os.Stdout
+	}
+	err := options.Stderr
+	if err == nil {
+		err = os.Stderr
+	}
```
mpl, line 237: it's a little bit inconsistent that all of this is done here before creating i, instead of being done after, with the rest of the options, i.e. L252, isn't it?

interp/interp.go
```diff
@@ -528,6 +549,62 @@ func (interp *Interpreter) Use(values Exports) {
 			interp.binPkg[k][s] = sym
 		}
 	}
+	if _, ok := values["fmt"]; ok {
+		fixStdio(interp)
+	}
```
mpl, line 554: what about the "log" package? what about any imported code that has its own printing functions/methods?
Ah, I see you're also hacking the log symbols L581. But are we only relying on the presence of "fmt" as a detection (L552) because you think any code that deals with input/output will always import "fmt" anyway?

interp/interp.go
```diff
@@ -547,8 +624,9 @@ func ignoreScannerError(e *scanner.Error, s string) bool {
 }
 
 // REPL performs a Read-Eval-Print-Loop on input reader.
-// Results are printed on output writer.
-func (interp *Interpreter) REPL(in io.Reader, out io.Writer) {
+// Results are printed on output writer. Errors are printed on errs writer.
```
mpl, line 627:
```suggestion
// Results are printed to the output writer of the Interpreter, provided as option
// at creation time. Errors are printed to the similarly defined errors writer.
```

interp/interp.go
```diff
@@ -207,16 +212,32 @@ func (n *node) Walk(in func(n *node) bool, out func(n *node)) {
 
 // Options are the interpreter options.
 type Options struct {
-	// GoPath sets GOPATH for the interpreter
+	// GoPath sets GOPATH for the interpreter.
 	GoPath string
-	// BuildTags sets build constraints for the interpreter
+	// BuildTags sets build constraints for the interpreter.
 	BuildTags []string
+	// Standard input, output and error streams.
```
mpl, line 219:
```suggestion
	// Standard input, output and error streams.
	// They default to os.Stdin, os.Stdout, and os.Stderr respectively.
```

interp/interp.go
```diff
@@ -614,16 +693,10 @@ func (interp *Interpreter) REPL(in io.Reader, out io.Writer) {
 		prompt(v)
 	}
 	cancel() // Do not defer, as cancel func may change over time.
+	return v, err
```
mpl, line 696: Maybe if err is nil at this point, we return s.Err() instead? But we have to see how that works out once this PR is rebased on top of master/PR 823.

cmd/yaegi/run.go
```diff
@@ -116,7 +116,7 @@ func runFile(i *interp.Interpreter, path string) error {
 	if s := string(b); strings.HasPrefix(s, "#!") {
 		// Allow executable go scripts, Have the same behavior as in interactive mode.
 		s = strings.Replace(s, "#!", "//", 1)
-		i.REPL(strings.NewReader(s), os.Stdout)
+		_, err = i.Eval(s)
```
mpl, line 119: Same question as for L59

cmd/yaegi/yaegi_test.go
```diff
@@ -108,8 +108,8 @@ func TestYaegiCmdCancel(t *testing.T) {
 			continue
 		}
 
-		if outBuf.String() != "context canceled\n" {
-			t.Errorf("unexpected output: %q", &outBuf)
+		if errBuf.String() != "context canceled\n" {
```
mpl, line 111:
```suggestion
		if strings.TrimSuffix(errBuf.String(), "\n") != context.Canceled.Error() {
```

mpl (changes requested):

cmd/yaegi/run.go
mvertes: Yes. Calling `Eval`directly here is simpler and clearer. Now we use `REPL` only in explicit interactive mode, which makes sense.

interp/interp.go
mvertes: ok.

interp/interp.go
mvertes: The call to `fixStdio()` needs to be done just after a `i.Use(stdlib.Symbols)`, and would be irrelevant for other arguments. Testing for a `"fmt"` package path achieves that. The only assumption here is that `"fmt"` is part of `"stdlib"`. I'm adding a comment to clarify that.

interp/interp.go
mvertes: ok

interp/interp.go
mvertes: ok

interp/interp.go
mvertes: Right.

cmd/yaegi/run.go
mvertes: Same answer as for L59

cmd/yaegi/yaegi_test.go
mvertes: ok

mpl (approved):

PR #836, opened 2020-09-02
fix: correct index for embedded binary method receiver

### LINKED ISSUE #834: panic due to wrong receiver value

```go
package main

import (
    "fmt"
    "io/ioutil"
    "log"
    "net/http"
    "net/http/httptest"
)

type T struct {
    http.ResponseWriter
}

type mw1 struct {
    next http.Handler
}   

func (m *mw1) ServeHTTP(rw http.ResponseWriter, rq *http.Request) {
    t := &T{
        ResponseWriter: rw,
    } 
    x := t.Header()
    fmt.Fprint(rw, "Welcome to my website!", x)
}   

func main() {
    m1 := &mw1{}
    
    mux := http.NewServeMux()
    mux.HandleFunc("/", m1.ServeHTTP)
    
    server := httptest.NewServer(mux)
    defer server.Close()

    client(server.URL)
}

func client(uri string) {
    resp, err := http.Get(uri)
    if err != nil {
        log.Fatal(err)
    }
    body, err := ioutil.ReadAll(resp.Body)
    if err != nil {
        log.Fatal(err)
    }
    fmt.Println(string(body))
}
```

Expected result:
```console
$ go run ./sample.go
Welcome to my website!map[]
```

Got:
```console
$ yaegi ./sample.go
unexpected fault address 0xc000436a20
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x2 addr=0xc000436a20 pc=0xc000436a20]

goroutine 5 [running]:
runtime.throw(0xed549c, 0x5)
        /usr/lib/go/src/runtime/panic.go:1116 +0x72 fp=0xc000201430 sp=0xc000201400 pc=0x43c792
runtime.sigpanic()
        /usr/lib/go/src/runtime/signal_unix.go:727 +0x405 fp=0xc000201460 sp=0xc000201430 pc=0x453465
runtime.call32(0xc000284210, 0xc000288078, 0xc0002860e0, 0x800000010)
        /usr/lib/go/src/runtime/asm_amd64.s:540 +0x3e fp=0xc000201490 sp=0xc000201460 pc=0x474cfe
reflect.Value.call(0xc0001e6750, 0xc000286040, 0x293, 0xed3bed, 0x4, 0x153d838, 0x0, 0x0, 0xc000286060, 0xec5de0, ...)
        /usr/lib/go/src/reflect/value.go:475 +0x8c7 fp=0xc0002016a8 sp=0xc000201490 pc=0x4d1a27
reflect.Value.Call(0xc0001e6750, 0xc000286040, 0x293, 0x153d838, 0x0, 0x0, 0xed11e0, 0xc0001e6701, 0x153d838)
        /usr/lib/go/src/reflect/value.go:336 +0xb9 fp=0xc000201728 sp=0xc0002016a8 pc=0x4d0f19
github.com/containous/yaegi/interp.callBin.func1(0xc0001e6750, 0xc000286040, 0x293, 0x153d838, 0x0, 0x0, 0x7b3195, 0xc0001e6750, 0xc000286040)
        /home/marc/go/src/github.com/containous/yaegi/interp/run.go:1085 +0x65 fp=0xc000201780 sp=0xc000201728 pc=0x7adaa5
[...]
```

mvertes:
When searching for a binary method on structures, look up on embedded
fields first, otherwise the resulting index is incorrect, as
reflect.Type.MethodByName succeeds also on container struct.

Fixes #834.

### DIFF
--- a/interp/type.go
+++ b/interp/type.go
@@ -1244,20 +1244,18 @@ func (t *itype) lookupBinMethod(name string) (m reflect.Method, index []int, isP
 	if t.cat == ptrT {
 		return t.val.lookupBinMethod(name)
 	}
+	for i, f := range t.field {
+		if f.embed {
+			if m2, index2, isPtr2, ok2 := f.typ.lookupBinMethod(name); ok2 {
+				index = append([]int{i}, index2...)
+				return m2, index, isPtr2, ok2
+			}
+		}
+	}
 	m, ok = t.TypeOf().MethodByName(name)
 	if !ok {
 		m, ok = reflect.PtrTo(t.TypeOf()).MethodByName(name)
 		isPtr = ok
-	}
-	if !ok {
-		for i, f := range t.field {
-			if f.embed {
-				if m2, index2, isPtr2, ok2 := f.typ.lookupBinMethod(name); ok2 {
-					index = append([]int{i}, index2...)
-					return m2, index, isPtr2, ok2
-				}
-			}
-		}
 	}
 	return m, index, isPtr, ok
 }
--- a/_test/cli6.go
+++ b/_test/cli6.go
@@ -1,0 +1,52 @@
+package main
+
+import (
+	"fmt"
+	"io/ioutil"
+	"log"
+	"net/http"
+	"net/http/httptest"
+)
+
+type T struct {
+	http.ResponseWriter
+}
+
+type mw1 struct {
+	next http.Handler
+}
+
+func (m *mw1) ServeHTTP(rw http.ResponseWriter, rq *http.Request) {
+	t := &T{
+		ResponseWriter: rw,
+	}
+	x := t.Header()
+	fmt.Fprint(rw, "Welcome to my website!", x)
+}
+
+func main() {
+	m1 := &mw1{}
+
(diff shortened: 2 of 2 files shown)


interp/type.go
```diff
@@ -1246,21 +1246,19 @@ func (t *itype) lookupBinMethod(name string) (m reflect.Method, index []int, isP
...
+				return m2, index, isPtr2, ok2
+			}
+		}
+	}
 	m, ok = t.TypeOf().MethodByName(name)
 	if !ok {
 		m, ok = reflect.PtrTo(t.TypeOf()).MethodByName(name)
 		isPtr = ok
 	}
-	if !ok {
```
mpl, line 1254: If above L1254, we just do:

```
	if ok {
		index = []int{m.Index}
	}
```
then the added test works too. So what is wrong with that solution?

mpl: answered by chat: it works by coincidence (m.Index is 0 too in this case), but m.Index does not have the meaning we want (i.e. it is not the position within the struct of the field that implements the method).

mpl (approved):

PR #838, opened 2020-09-08
feature: test subcommand to run test and benchmark functions

mvertes:
This change allows the interpreter to execute tests and benchmarks
functions provided by packages.

The test subcommand is similar to the "go test" command and
all the relevant flags have been kept.

The ability to evaluate a directory or a package has also been added.

A new method Symbol to access exported symbol values of an interpreted
package has been added. This method is used by the test subcommand.

An EvalTest method has been added to evaluate all Go files, including "*_test.go".

The testing packages from the standard library have been added to stdlib used
symbols.


cmd/yaegi/test.go
```diff
@@ -0,0 +1,132 @@
...
+		return err
+	}
+	args := tflag.Args()
+	path := "."
+	if len(args) > 0 {
+		path = args[0]
+	}
+
+	// Overwrite os.Args with correct flags to setup testing.Init.
+	tf := []string{""}
```
mpl, line 67: Do we actually want an empty string as the first element of the slice?

cmd/yaegi/test.go
```diff
@@ -0,0 +1,132 @@
...
+		switch fun := sym.Interface().(type) {
+		case func(*testing.B):
+			benchmarks = append(benchmarks, testing.InternalBenchmark{name, fun})
+		case func(*testing.T):
+			tests = append(tests, testing.InternalTest{name, fun})
+		}
+	}
+
+	testing.Main(regexp.MatchString, tests, benchmarks, nil)
+	return err
```
mpl, line 131: That is always nil, isn't it?

interp/build_test.go
```diff
@@ -74,7 +74,7 @@ func TestBuildFile(t *testing.T) {
 	for _, test := range tests {
 		test := test
 		t.Run(test.src, func(t *testing.T) {
-			if r := skipFile(&ctx, test.src); r != test.res {
+			if r := skipFile(&ctx, test.src, true); r != test.res {
```
mpl, line 77:
```suggestion
			if r := skipFile(&ctx, test.src, NoTest); r != test.res {
```
?

interp/gta.go
```diff
@@ -216,7 +216,7 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
 					err = n.cfgErrorf("%s redeclared in this block", name)
 					return false
 				}
-			} else if pkgName, err = interp.importSrc(rpath, ipath); err == nil {
+			} else if pkgName, err = interp.importSrc(rpath, ipath, true); err == nil {
```
mpl, line 219:
```suggestion
			} else if pkgName, err = interp.importSrc(rpath, ipath, NoTest); err == nil {
```

interp/interp.go
```diff
@@ -361,30 +366,64 @@ func (interp *Interpreter) resizeFrame() {
...
-	// TODO(marc): implement eval of a directory, package and tests.
-	b, err := ioutil.ReadFile(path)
-	if err != nil {
-		return res, err
+// EvalPath evaluates Go code located at path, running tests specified by tfe
+// which is the test filter regular expression. If tfe is empty, the main function
+// is evaluated. EvalPath returns the last result computed by the interpreter,
+// and a non nil error in case of failure.
+func (interp *Interpreter) EvalPath(path string, skiptest bool) (res reflect.Value, err error) {
+	if isFile(path) {
```
mpl, line 380: Maybe invert that test, for less indented code?

mpl (commented):
First pass, next pass tomorrow.

interp/interp.go
```diff
@@ -439,25 +478,26 @@ func (interp *Interpreter) eval(src, name string, inc bool) (res reflect.Value,
...
-
 	if root.kind != fileStmt {
 		// REPL may skip package statement
 		setExec(root.start)
 	}
 	interp.mutex.Lock()
+	gs := interp.scopes[pkgName]
 	if interp.universe.sym[pkgName] == nil {
 		// Make the package visible under a path identical to its name
 		// TODO(mpl): srcPkg is supposed to be keyed by importPath. Verify it is necessary, and implement.
```
mpl, line 489: we can remove that comment, it's already done.

interp/interp.go
```diff
@@ -439,25 +478,26 @@ func (interp *Interpreter) eval(src, name string, inc bool) (res reflect.Value,
...
+		interp.srcPkg[pkgName] = gs.sym
 		interp.universe.sym[pkgName] = &symbol{kind: pkgSym, typ: &itype{cat: srcPkgT, path: pkgName}}
 		interp.pkgNames[pkgName] = pkgName
 	}
 	interp.mutex.Unlock()
 
+	// Add main to list of functions to run, after all inits
+	if m := gs.sym[mainID]; m != nil {
+		initNodes = append(initNodes, m.node)
+	}
```
mpl, line 499: I don't think this is correct. I think we only want to add the main function to the scope that is called "main", right? But the way you have defined gs L486 does not guarantee that gs is main scope, since pkgName could be anything.
I think I had added that extra check at some point, because when inspecting the whole symbols map I was seeing extra "main" symbols that should not exist.

interp/interp.go
```diff
@@ -361,30 +366,64 @@ func (interp *Interpreter) resizeFrame() {
...
-func (interp *Interpreter) EvalPath(path string) (res reflect.Value, err error) {
-	// TODO(marc): implement eval of a directory, package and tests.
-	b, err := ioutil.ReadFile(path)
-	if err != nil {
-		return res, err
+// EvalPath evaluates Go code located at path, running tests specified by tfe
+// which is the test filter regular expression. If tfe is empty, the main function
+// is evaluated. EvalPath returns the last result computed by the interpreter,
+// and a non nil error in case of failure.
+func (interp *Interpreter) EvalPath(path string, skiptest bool) (res reflect.Value, err error) {
```
mpl, line 379: I'm not super happy about the extra boolean everywhere obviously, but there does not seem to be any better alternative. :( 
Maybe for once we could consider a variadic...

interp/src.go
```diff
@@ -128,32 +128,33 @@ func (interp *Interpreter) importSrc(rPath, importPath string) (string, error) {
...
 	if err != nil {
 		return "", err
 	}
 	interp.run(n, nil)
 
-	// Add main to list of functions to run, after all inits
-	if m := interp.main(); m != nil {
-		initNodes = append(initNodes, m)
+	// Add main to list of functions to run, after all inits.
+	if m := gs.sym[mainID]; m != nil && skipTest {
```
mpl, line 156: Same remark as above. I'm not sure this is correct, as importPath may be something other than "main".

interp/interp.go
```diff
@@ -361,30 +366,64 @@ func (interp *Interpreter) resizeFrame() {
...
+		case constSym:
+			m[n] = s.rval
+		case funcSym:
+			m[n] = genFunctionWrapper(s.node)(interp.frame)
+		case varSym:
+			if i := s.index; i >= 0 {
+				m[n] = interp.frame.data[i]
+			} else {
+				m[n] = genValue(s.node)(interp.frame)
+			}
```
mpl, line 418: as discussed by chat, this might not be needed

cmd/yaegi/test.go
mvertes: `flag` will parse from `os.Args[1:]`, so the content of first element is ignored, no need to set it to a particular value.

cmd/yaegi/test.go
mvertes: yes.

interp/build_test.go
mvertes: ok

interp/gta.go
mvertes: ok

interp/interp.go
mvertes: if will replace L497 by:
```go
if m := gs.sym[mainID]; pkgName == mainID && m != nil {
```
to add the main symbol only if in a main packge

interp/interp.go
mvertes: The alternate is to keep `EvalPath(path)` as is and add `EvalTestPath(path)`.

cmd/yaegi/test.go
mpl: Ah, I did not see that os.Args becomes tf. That is why we want the flags to start at tf[1], and not tf[0], ok.

interp/interp.go
mpl: yes, much better imho.

interp/interp.go
mpl: as a tiny optimization, maybe it's faster to test pkgName == mainID before doing the rest of the test, but I'm not sure.

interp/src.go
mvertes: will replace L156 by:
```go
if m := gs.sym[mainID]; pkgName == mainID && m != nil && skipTest {
```

interp/interp.go
mvertes: ok

interp/interp.go
mvertes: This test is performed only once per call to `eval` or to `importSrc`. I would favor conciseness and readability here.

_test/m1/main_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"fmt"
+	"math/rand"
+	"testing"
+)
+
+func TestMain(t *testing.T) {
+	fmt.Println("in test")
+	//t.Errorf("This is test")
```
mpl, line 11: remove?

cmd/yaegi/test.go
```diff
@@ -0,0 +1,131 @@
...
+	for name, sym := range i.Symbols(path) {
+		switch fun := sym.Interface().(type) {
+		case func(*testing.B):
+			benchmarks = append(benchmarks, testing.InternalBenchmark{name, fun})
+		case func(*testing.T):
+			tests = append(tests, testing.InternalTest{name, fun})
+		}
+	}
+
+	testing.Main(regexp.MatchString, tests, benchmarks, nil)
```
mpl, line 129: I wonder if maybe the interp API should provide a convenience method that does this part?

Also, here's a weird edge case: what if someone writes a func(*testing.B) in a .go file that is _not_ a *_test.go file? then this function will end up in the list of tests to run, right? If yes, then:
1) How does the go tool handle that?
2) How do we want to handle it?

mvertes: At this stage I prefer not to change the interp API, due to the fact that `testing.Main()` is calling `os.Exit()`. In order to make `testing` part of the API, we would have to use directly `MainStart` method of `testing.M`, not guaranteed to be stable across versions, and still have to cope with the strange way of setting options through `flag` and global `os.Args`. Definitely not perfect.

I added a test directly in `_test/m1/main.go` in order to respond to your questions:

`go run` and `yaegi run` both do not execute the test, as expected.
`go test` do not execute the test part of main, `yaegi test` does. Both options are acceptable to me. I do not think it is formally specified.

mpl (approved):

PR #849, opened 2020-09-14
interp: fix data race for composite literal creation

mpl:
This change fixes two data races related to composite literal creation.

The first one isn't controversial as it is just about initializing the
variable that contains the values in the right place, i.e. within the
n.exec, so that this variable is local to each potential goroutine,
instead of being racily shared by all goroutines.

The second one is more worrying, i.e. having to protect the node typ
with a mutex, because a call to func (t *itype) refType actually
modifies the itype itself, which means it is not concurrent safe.
The change seems to work, and does not seem to introduce regression, but
it is still a concern as it probably is a sign that more similar
guarding has to be done in several other places.

### DIFF
--- a/interp/interp_eval_test.go
+++ b/interp/interp_eval_test.go
@@ -1111,6 +1111,51 @@ func TestConcurrentEvals3(t *testing.T) {
 	}
 }
 
+func TestConcurrentComposite1(t *testing.T) {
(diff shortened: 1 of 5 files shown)


mvertes (approved):
LGTM

PR #855, opened 2020-09-18
chore: adds a downloader script.

ldez:
The goal is to be able to easily use the Yaegi command line in a CI.
This will be very useful for Traefik plugins creators, or for anyone who wants to use an interpreted script in a CI.


mvertes (changes requested):
I do not understand exactly why this `godownloader.sh` script is required. Why do we need to include it ?

May be some use instructions need to be added in `README.md` ?

It could be better to rename `godownloader.sh` in `install.sh`, for clarity, as in http://golangci/golangci-lint.
As this script is itself generated, a rule to rebuild it (if we change the list of supported archictectures) needs to be added in `Makefile` .

ldez:
The goal is to be able to easily use the Yaegi command line in a CI.
This will be very useful for Traefik plugins creators, or for anyone who wants to use an interpreted script in a CI.

mvertes (commented):
Ok, thanks. LGTM.

mvertes (approved):

PR #861, opened 2020-09-22
interp: refactor doComposite cases

### LINKED ISSUE #851: interp: wrong value for composite literal initialization
Note how this is the same test as _test/composite8.go, but with the key specified for the composite literal.

```go
package main

type T struct{ I int }

func main() {
	t := []*T{}
	s := []int{1, 2}
	for _, e := range s {
		x := &T{I: e}
		t = append(t, x)
	}
	println(t[0].I, t[1].I)
}

// Output:
// 1 2
```

Expected result:
```console
$ go run ./sample.go
// 1 2
```

Got:
```console
$ yaegi ./sample.go
// 2 2
```

### LINKED ISSUE #854: interp: panic when assigning struct field as composite literal in non-keyed form

```go
package main

 import "fmt"

type A struct {
	C D
}

type D struct {
	E string
}

func main() {
/*
	// Note that this works fine
	// (direct initialization of field C)
	a := &A{
		C: D{"bb"},
	}
	// And so does this
	// (keyed form for the E field of D)
	a := &A{}
	a.C = D{E: "bb"}
*/
	// But this does not
	// (non-keyed form for the E field of D)
	a := &A{}
	a.C = D{"bb"}
	fmt.Println(a.C.E)
}
```

Expected result:
```console
$ go run ./sample.go
bb
```

Got:
```console
$ yaegi ./sample.go
reflect: call of reflect.Value.Field on zero Value
goroutine 1 [running]:
runtime/debug.Stack(0x1, 0xc0002e2800, 0x40)
	/Users/mpl/go1/src/runtime/debug/stack.go:24 +0x9d
github.com/containous/yaegi/interp.(*Interpreter).eval.func1(0xc0000b3d08)
	/Users/mpl/src/github.com/containous/yaegi/interp/interp.go:451 +0xbc
panic(0x198a0c0, 0xc000209660)
	/Users/mpl/go1/src/runtime/panic.go:967 +0x15d
github.com/containous/yaegi/interp.runCfg.func1(0xc00036a0a0, 0xc000470d00)
	/Users/mpl/src/github.com/containous/yaegi/interp/run.go:124 +0x20c
panic(0x198a0c0, 0xc000209660)
	/Users/mpl/go1/src/runtime/panic.go:967 +0x15d
reflect.Value.Field(0x0, 0x0, 0x0, 0x0, 0xc00020a328, 0xc0002095d0, 0xc0000b3a70)
	/Users/mpl/go1/src/reflect/value.go:827 +0x123
reflect.Value.FieldByIndex(0x0, 0x0, 0x0, 0xc00047a8c8, 0x1, 0x1, 0xc000210b80, 0xc0002095d0, 0x196)
	/Users/mpl/go1/src/reflect/value.go:859 +0x2c2
github.com/containous/yaegi/interp.getPtrIndexSeq.func3(0xc00036a0a0, 0xc000210cc0)
	/Users/mpl/src/github.com/containous/yaegi/interp/run.go:1620 +0xdf
github.com/containous/yaegi/interp.runCfg(0xc000470d00, 0xc00036a0a0)
	/Users/mpl/src/github.com/containous/yaegi/interp/run.go:130 +0x6d
github.com/containous/yaegi/interp.(*Interpreter).run(0xc000364000, 0xc000470200, 0xc00036a000)
	/Users/mpl/src/github.com/containous/yaegi/interp/run.go:108 +0x296
github.com/containous/yaegi/interp.(*Interpreter).eval(0xc000364000, 0xc0002e5680, 0x205, 0x7ffeefbff68a, 0x11, 0x200, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/mpl/src/github.com/containous/yaegi/interp/interp.go:543 +0x7d7
github.com/containous/yaegi/interp.(*Interpreter).EvalPath(0xc000364000, 0x7ffeefbff68a, 0x11, 0x400, 0xc0002e5440, 0x205, 0x0, 0xc000216c30)
	/Users/mpl/src/github.com/containous/yaegi/interp/interp.go:392 +0x102
main.runFile(0xc000364000, 0x7ffeefbff68a, 0x11, 0xc0002062a0, 0x1)
	/Users/mpl/src/github.com/containous/yaegi/cmd/yaegi/run.go:113 +0x158
main.run(0xc0000321d0, 0x1, 0x1, 0x1aca7e8, 0x1967740)
	/Users/mpl/src/github.com/containous/yaegi/cmd/yaegi/run.go:77 +0x708
main.main()
	/Users/mpl/src/github.com/containous/yaegi/cmd/yaegi/yaegi.go:119 +0x2f4

run: reflect: call of reflect.Value.Field on zero Value
```

This is all related to the fact that the non-keyed form of a composite literal is initialized with `getFrame(f, l).data[frameIndex] = a` , while the keyed form is with `d.Set(a)` (which is actually incorrect too, see #851 ).

Edit: the problem is also somehow related to the "writability" of the A.C field. The observation hinting in that direction if that if we instead have:

```
type A struct {
	C *D
}
```

then the issue does not manifest itself.

mpl:
Also, there is an optimisation that used to take place when the RHS of
an assignment is a struct literal. That optimisation avoids some
allocation by "instructing" that whatever we want to write for the frame
of the RHS gets directly written to the frame of the LHS.
But it just so happens that we're doing this incorrectly when the LHS is
a struct field (or probably for any other kind of getIndex AST
action). Therefore this PR also disables the optimisation when the LHS
is a getIndex.

Fixes #851
Fixes #854

### DIFF
--- a/_test/composite12.go
+++ b/_test/composite12.go
@@ -1,0 +1,19 @@
+package main
+
+type A struct {
+	C D
+}
+
+type D struct {
+	E string
+}
+
+func main() {
+	a := A{}
+	a.C = D{"bb"}
+
+	println(a.C.E)
+}
+
+// Output:
+// bb
--- a/_test/composite13.go
+++ b/_test/composite13.go
@@ -1,0 +1,19 @@
+package main
+
+type A struct {
+	C D
+}
+
+type D struct {
+	E string
+}
+
+func main() {
+	a := A{}
+	a.C = D{E: "bb"}
+
+	println(a.C.E)
+}
+
+// Output:
+// bb
--- a/_test/composite8bis.go
+++ b/_test/composite8bis.go
@@ -1,0 +1,16 @@
+package main
+
+type T struct{ I int }
+
+func main() {
+	t := []*T{}
+	s := []int{1, 2}
+	for _, e := range s {
+		x := &T{I: e}
+		t = append(t, x)
+	}
+	println(t[0].I, t[1].I)
+}
+
+// Output:
+// 1 2
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -543,6 +543,12 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 						// which require and additional operation to set the value
 						break
 					}
+					if dest.action == aGetIndex {
+						// optimization does not work when assigning to a struct field. Maybe we're not
+						// setting the right frame index or something, and we would end up not writing at
+						// the right place. So disabling it for now.
+						break
+					}
 					// Skip the assign operation entirely, the source frame index is set
 					// to destination index, avoiding extra memory alloc and duplication.
 					n.gen = nop
@@ -2368,9 +2374,9 @@ func compositeGenerator(n *node, typ *itype) (gen bltnGenerator) {
 			gen = compositeLitNotype
 		case n.lastChild().kind == keyValueExpr:
 			if n.nleft == 1 {
-				gen = compositeSparse
+				gen = compositeLitKeyed
 			} else {
-				gen = compositeSparseNotype
+				gen = compositeLitKeyedNotype
 			}
 		default:
 			if n.nleft == 1 {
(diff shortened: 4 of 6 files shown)


mvertes (approved):
LGTM

PR #869, opened 2020-09-27
fix: yaegi os.Args should contain the script name

### LINKED ISSUE #868: flag.Parse() causes panic

```go
#!/usr/bin/env yaegi
package main

import "flag"

func main() {
	var n = flag.Int("i", 1, "provide an int")
	flag.Parse()
}
```

Expected result:
No output

Got:
```console
$ ./sample.go
10:10: panic
runtime error: slice bounds out of range [1:0]
goroutine 1 [running]:
[...]
run: runtime error: slice bounds out of range [1:0]
```

nrwiersma:
`flag.Parse()` fails with a panic when run using `yaegi run` as there are an unexpected number of args. It expects the first arg to be the application name. In the case of yaegi it makes sense that this is the script name. 

Fixes #868

### DIFF
--- a/cmd/yaegi/run.go
+++ b/cmd/yaegi/run.go
@@ -70,7 +70,7 @@ func run(arg []string) error {
 
 	// Skip first os arg to set command line as expected by interpreted main
 	path := args[0]
-	os.Args = arg[1:]
+	os.Args = arg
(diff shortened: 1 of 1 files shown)


mvertes (approved):
LGTM

PR #874, opened 2020-09-28
fix: convert struct tags properly

### LINKED ISSUE #873: struct tags are not converted from constant values

```go
package main

import (
	"fmt"
	"reflect"
)

type A struct {
	Test string `tag:"test"`
}

func main() {
	a := A{}
	t := reflect.TypeOf(a)
	f, ok := t.FieldByName("Test")
	if !ok {
		return
	}

	fmt.Println(f.Tag.Get("tag"))
}
```

Expected result:
```console
$ go run ./sample.go
test
```

Got:
```console
$ yaegi ./sample.go

```

nrwiersma:
Struct tags were forgotten in the conversion of constant values. They need to be converted from constant values before being assigned.

Fixes #873

### DIFF
--- a/_test/struct58.go
+++ b/_test/struct58.go
@@ -1,0 +1,24 @@
+package main
+
+import (
+	"fmt"
+	"reflect"
+)
+
+type A struct {
+	Test string `tag:"test"`
+}
+
+func main() {
+	a := A{}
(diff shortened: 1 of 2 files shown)


mvertes (approved):
LGTM

PR #876, opened 2020-09-29
fix: compute array type size from constant expression

### LINKED ISSUE #875: create array with size from forward defined const

```go
package main

import "fmt"

type T [l1 + l2]int

const (
    l1 = 2
    l2 = 3
)

func main() {
    fmt.Println(T{})
}
```

Expected result:
```console
$ go run ./sample.go
[0 0 0 0 0]
```

Got:
```console
$ yaegi ./sample.go
./sample.go:5:9: undefined: l1
```
This bug is related to the issue #865, although there are may be others problems lurking behind in this issue.

mvertes:
Fixes #875.


interp/type.go
```diff
@@ -198,11 +198,16 @@ func nodeType(interp *Interpreter, sc *scope, n *node) (*itype, error) {
...
-					if _, err = interp.cfg(n.child[0], sc.pkgID); err != nil {
+					if _, err = interp.cfg(c0, sc.pkgID); err != nil {
 						return nil, err
 					}
-					t.incomplete = true
+					if v, ok := c0.rval.Interface().(constant.Value); ok {
+						t.size = constToInt(v)
+					} else {
+						t.incomplete = true
+					}
```
mpl, line 210: We are in a switch, so we can break out early on error, like so:

```suggestion
					if v, ok := c0.rval.Interface().(constant.Value); !ok {
						t.incomplete = true
						break
					}
					t.size = constToInt(v)
```

mpl (changes requested):

interp/type.go
```diff
@@ -186,14 +186,14 @@ func nodeType(interp *Interpreter, sc *scope, n *node) (*itype, error) {
...
-							t.size = constToInt(c)
-						} else {
-							t.incomplete = true
-						}
+					if sym.typ == nil || sym.typ.cat != intT {
+						t.incomplete = true
+						break
+					}
+					if v, ok := sym.rval.Interface().(int); ok {
+						t.size = v
```
mpl, line 194: you can also break after this one, then "else if" below simply becomes an "if".

interp/type.go
```diff
@@ -186,14 +186,14 @@ func nodeType(interp *Interpreter, sc *scope, n *node) (*itype, error) {
...
+					} else if c, ok := sym.rval.Interface().(constant.Value); ok {
+						t.size = constToInt(c)
```
mpl, line 196: same here, and then the "else" below can be removed.

interp/type.go
```diff
@@ -166,56 +166,65 @@ func nodeType(interp *Interpreter, sc *scope, n *node) (*itype, error) {
...
-				// constant size
-				if isConstantValue(v.Type()) {
-					c := v.Interface().(constant.Value)
-					t.size = constToInt(c)
-				} else {
-					t.size = int(v.Int())
+		c0 := n.child[0]
+		if len(n.child) == 1 {
+			// Array size is not defined.
+			if t.val, err = nodeType(interp, sc, c0); err != nil {
```
mpl, line 172: This change has a side-effect: previously, we would do `if t.val, err = nodeType(interp, sc, n.child[0])` even when `len(n.child) == 0`. So I suppose if that happened for some unforeseen reason, we would get a panic and we would at least know we made a mistake somewhere.
With the new code, the case `len(n.child) == 0` is completely skipped. I don't know whether we should be super defensive and add an explicit panic or something in case that happens. I leave that to your appreciation.

mpl (approved):
Much cleaner now, thanks.

interp/type.go
mvertes: `len(n.child[0])` means here an array declaration with no type, which I think is not possible in this context, and would have been catch earlier by the Go parser.

PR #877, opened 2020-09-29
fix: replace playground link in docs

JulienBreux:
### Context

Replacing playground link in doc, the previous code imported containous instead of traefik.

### ToDo

- [x] Create new playground code with the right `import`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -74,7 +74,7 @@ func main() {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mvertes (approved):
Looks good to me.

mvertes:
Thanks for the correction!

PR #881, opened 2020-09-30
interp: apply integer division when appropriate

### LINKED ISSUE #864: REPL: type conversion/comparison error after division

```console
% yaegi
> 1/1
1 not an Int
goroutine 7 [running]:
runtime/debug.Stack(0x1, 0xc00016a400, 0x40)
	/Users/mpl/go1/src/runtime/debug/stack.go:24 +0x9d
github.com/traefik/yaegi/interp.(*Interpreter).eval.func1(0xc000177f78)
	/Users/mpl/src/github.com/traefik/yaegi/interp/interp.go:451 +0xbc
panic(0x194eae0, 0xc000196540)
	/Users/mpl/go1/src/runtime/panic.go:967 +0x15d
go/constant.Int64Val(0x1c93520, 0xc000045a00, 0x194dd20, 0x1c93520)
	/Users/mpl/go1/src/go/constant/value.go:499 +0x19d
github.com/traefik/yaegi/interp.convertConstantValue(0xc000140a00)
	/Users/mpl/src/github.com/traefik/yaegi/interp/run.go:2825 +0x66d
github.com/traefik/yaegi/interp.genValue(0xc000140a00, 0x0)
	/Users/mpl/src/github.com/traefik/yaegi/interp/value.go:116 +0x443
github.com/traefik/yaegi/interp.(*Interpreter).eval(0xc000362000, 0xc0000368d8, 0x4, 0x0, 0x0, 0x1, 0x0, 0x0, 0x0, 0x0, ...)
	/Users/mpl/src/github.com/traefik/yaegi/interp/interp.go:545 +0x7fa
github.com/traefik/yaegi/interp.(*Interpreter).Eval(...)
	/Users/mpl/src/github.com/traefik/yaegi/interp/interp.go:376
github.com/traefik/yaegi/interp.(*Interpreter).EvalWithContext.func1(0xc000042300, 0xc000362000, 0xc0000368d8, 0x4, 0xc00000ed80, 0xc000196270)
	/Users/mpl/src/github.com/traefik/yaegi/interp/interp.go:572 +0x84
created by github.com/traefik/yaegi/interp.(*Interpreter).EvalWithContext
	/Users/mpl/src/github.com/traefik/yaegi/interp/interp.go:570 +0x14d
```

Expected result:
```console
1
```

N.B: this works fine in non-interactive mode.

mpl:
When working with an untyped const expression involving a division, if
the default type of the result should be an int (for example because the
default types of all the operands are ints as well), then we should make
sure that the operation that is applied is indeed an integer division,
and that the type of the result is not a float.

This is achieved by using the QUO_ASSIGN operator, instead of the QUO
operator.

This should fix several problems lurking around, and it notably fixes
one of the visible consequences, which is a systematic panic when using
the REPL as a "calculator".

This incidentally also allows us to revert what was done in
5dfc3b86dc7753b035343261339663dd78520412 since it now turns out it was
just a hack to fix one of the symptoms.

Fixes #864

### DIFF
--- a/internal/cmd/genop/genop.go
+++ b/internal/cmd/genop/genop.go
@@ -197,8 +197,10 @@ func {{$name}}Const(n *node) {
 		v := constant.Shift(vConstantValue(v0), token.{{tokenFromName $name}}, uint(vUint(v1)))
 		n.rval.Set(reflect.ValueOf(v))
 		{{- else if (eq $op.Name "/")}}
-		// TODO(mpl): exclude uints?
 		var operator token.Token
+		// When the result of the operation is expected to be an int (because both
+		// operands are ints), we want to force the type of the whole expression to be an
+		// int (and not a float), which is achieved by using the QUO_ASSIGN operator.
 		if n.typ.untyped && isInt(n.typ.rtype) {
 			operator = token.QUO_ASSIGN
 		} else {
--- a/internal/cmd/genop/genop.go
+++ b/internal/cmd/genop/genop.go
@@ -196,6 +196,16 @@ func {{$name}}Const(n *node) {
 		{{- if $op.Shift}}
 		v := constant.Shift(vConstantValue(v0), token.{{tokenFromName $name}}, uint(vUint(v1)))
 		n.rval.Set(reflect.ValueOf(v))
+		{{- else if (eq $op.Name "/")}}
+		// TODO(mpl): exclude uints?
+		var operator token.Token
+		if n.typ.untyped && isInt(n.typ.rtype) {
+			operator = token.QUO_ASSIGN
+		} else {
+			operator = token.QUO
+		}
+		v := constant.BinaryOp(vConstantValue(v0), operator, vConstantValue(v1))
+		n.rval.Set(reflect.ValueOf(v))
 		{{- else}}
 		v := constant.BinaryOp(vConstantValue(v0), token.{{tokenFromName $name}}, vConstantValue(v1))
 		n.rval.Set(reflect.ValueOf(v))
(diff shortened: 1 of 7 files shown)


internal/cmd/genop/genop.go
```diff
@@ -196,6 +196,16 @@ func {{$name}}Const(n *node) {
 		{{- if $op.Shift}}
 		v := constant.Shift(vConstantValue(v0), token.{{tokenFromName $name}}, uint(vUint(v1)))
 		n.rval.Set(reflect.ValueOf(v))
+		{{- else if (eq $op.Name "/")}}
+		// TODO(mpl): exclude uints?
```
mvertes, line 200: As the default type for untyped integer constants it `int`, nothing needs to be done with regard to uints. Your comment can be disgarded.

internal/cmd/genop/genop.go
```diff
@@ -196,6 +196,16 @@ func {{$name}}Const(n *node) {
...
+		var operator token.Token
+		if n.typ.untyped && isInt(n.typ.rtype) {
+			operator = token.QUO_ASSIGN
```
mvertes, line 203: Can you add a comment explaining that `token.QUO_ASSIGN` forces a integer division, as a special case in `constant` package, as it is not obvious when reading the code ?

mvertes (changes requested):

internal/cmd/genop/genop.go
mpl: done

mvertes (approved):
LGTM

PR #891, opened 2020-10-07
fix: allow yaegi command to interpret itself

### LINKED ISSUE #890: yaegi interpreting itself

```console
$ yaegi  -syscall -unsafe -unrestricted github.com/traefik/yaegi/cmd/yaegi -e "println(1)"
```

Expected result:
```console
1
```

Got:
```console
cmd/yaegi/run.go:122:19: package interp "github.com/traefik/yaegi/interp" has no symbol Panic
run: cmd/yaegi/run.go:122:19: package interp "github.com/traefik/yaegi/interp" has no symbol Panic
```

mvertes:
Since the introduction of restricted stdlib and syscall symbols, the
capability of yaegi to interpret itself was broken.
The use of unrestricted symbols is now also controlled by environment
variables, to allow propagation accross nested interpreters.
The interpreter Panic symbol was not wrapped, this is fixed now.
the import path resolution was failing if the working directory was
outside of GOPATH.
The documentation and readme have been ajusted.

Fixes #890.

### DIFF
--- a/cmd/yaegi/run.go
+++ b/cmd/yaegi/run.go
@@ -6,6 +6,7 @@ import (
 	"go/build"
 	"io/ioutil"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/traefik/yaegi/interp"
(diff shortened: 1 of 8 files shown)


interp/run.go
```diff
@@ -2034,7 +2034,7 @@ func doCompositeBinStruct(n *node, hasType bool) {
 			}
 		} else {
 			fieldIndex[i] = []int{i}
-			if c.typ.cat == funcT {
+			if c.typ.cat == funcT && len(c.child) > 1 {
```
mpl, line 2037: I do not understand this change.
I understand we are in the case of a literal composite, and one of its fields is defined without a key, just the value. and the value is a function. Could you give me an example of a case where
 `c.typ.cat == funcT && len(c.child) == 1`
and one where
 `c.typ.cat == funcT && len(c.child) == 2`
so i can understand the difference?

mpl (changes requested):

cmd/yaegi/run.go
```diff
@@ -46,13 +49,22 @@ func run(arg []string) error {
 	i.Use(interp.Symbols)
 	if useSyscall {
 		i.Use(syscall.Symbols)
+		if err := os.Setenv("YAEGI_SYSCALL", "1"); err != nil {
+			return err
+		}
```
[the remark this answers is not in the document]
mvertes: That's right. Will add a comment about that.

interp/run.go
mvertes: This may happen when initializing a runtime defined struct with a function field (`typ.Field(i).Type == reflect.Func`) with an interpreter defined stuct with a function field already being a reflect.Value from runtime.

I met this case when interpreting (inception mode) `cmd/yaegi/test.go` at line 123:
```go
benchmarks = append(benchmarks, testing.InternalBenchmark{name, fun})
```
where `fun` is provided through a type switch assign, thus already converted to the runtime type.

mpl: something's still unclear to me, but I'll clarify with you later. No need to block the PR for that.

mpl (approved):

PR #894, opened 2020-10-09
fix: correct handling of select comm clause with empty body

### LINKED ISSUE #887: Select statement does no proceed if no body statements given
The following program `sample.go` demonstrates the problem:

```go
package main

import (
	"time"
)

func main() {
	t := time.NewTicker(time.Second)
	defer t.Stop()

	ch := make(chan struct{})
	defer close(ch)

	for i := 0; i < 3; i++ {
		println("here")

		select {
		case <-ch:
			return
		case <-t.C:
		}
	}
}
```

Expected result:
```console
$ go run ./sample.go
here
here
here
```

Got:
```console
$ yaegi ./sample.go
here
```

mvertes:
Fixes #887.

### DIFF
--- a/_test/select14.go
+++ b/_test/select14.go
@@ -1,0 +1,23 @@
+package main
+
+type T struct {
+	c1 chan string
+	c2 chan string
+}
+
+func main() {
+	t := &T{}
+	t.c2 = make(chan string)
+
+	go func(c chan string) { c <- "done" }(t.c2)
+
+	select {
+	case msg := <-t.c1:
+		println("received from c1:", msg)
+	case <-t.c2:
+	}
+	println("Bye")
+}
+
+// Output:
+// Bye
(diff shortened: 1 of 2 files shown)


_test/select14.go
```diff
@@ -0,0 +1,23 @@
...
+
+func main() {
+	t := &T{}
+	t.c2 = make(chan string)
+
+	go func(c chan string) { c <- "done" }(t.c2)
+
+	select {
+	case msg := <-t.c1:
+		println("received from c1:", msg)
```
mpl, line 16: Is this case useful for the test?

mvertes: It makes sense to have multiple comm clauses to test for correct branching, at CFG level and at run level.

mpl (approved):

PR #896, opened 2020-10-09
fix: correctly return constant expressions in functions

### LINKED ISSUE #889: Functions can't return time.Duration multiplied by constant
The following program `sample.go` has an error:

```go
package main

import (
	"fmt"

)

const script = `
package bar

import "time"

func get10Hours() time.Duration {
	return 10*time.Hour
}

func Bar() string {
	return "ok, " + get10Hours().String()
}
`

func main() {
	i.Use(stdlib.Symbols)

	_, err := i.Eval(script)
	handle(err)

	v, err := i.Eval("bar.Bar")
	handle(err)

	bar := v.Interface().(func() string)

	result := bar()
	fmt.Println(result)
}

func handle(err error) {
	if err != nil {
		panic(err)
	}
}
```

Expected result:
```console
ok, 10h0m0s
```

Got:
```console
ok, 0s
```

---

is you change script to:
```go
const script = `
package bar

import "time"

func getHour() time.Duration {
	return time.Hour
}

func Bar() string {
	return "ok, " + getHour().String()
}
`
```

then result will be:
```console
ok, 1h0m0s
```

As expected

mvertes:
In the case of a function returning a single value, a constant
result could be ignored, and the function would return a zero value.

Fixes #889.


interp/run.go
```diff
@@ -1869,7 +1869,7 @@ func _return(n *node) {
 	case 0:
 		n.exec = nil
 	case 1:
-		if child[0].kind == binaryExpr || isCall(child[0]) {
+		if !child[0].rval.IsValid() && child[0].kind == binaryExpr || isCall(child[0]) {
```
mpl, line 1872:
```suggestion
		// This is an optimisation that is applied for binary expressions or function
		// calls, but not for (binary) expressions involving const, as the values are not
		// stored in the frame in that case.
		if !child[0].rval.IsValid() && child[0].kind == binaryExpr || isCall(child[0]) {
```

mpl (changes requested):

mpl (approved):

PR #898, opened 2020-10-10
feature: provide a version sub-command

### LINKED ISSUE #840: yaegi: new "version" subcommand
#### Proposal

`yaegi version` should print the version of the code that was used to build the current yaegi binary.
In the case we're not on an exact semver, if possible, the git commit could be printed.

#### Background

It would particularly help for bug reports, to make it easier for users to report which version they are using.

mvertes:
A version sub-command is implemented.
A local install target is added to Makefile, to set version from git at build.

Fixes #840.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ matrix:
 env:
   global:
     - GO111MODULE=on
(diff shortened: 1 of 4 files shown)


ldez:
https://goreleaser.com/customization/build/

ldez (approved):
LGTM

PR #909, opened 2020-10-16
interp: fix assignable check

mpl:
The assignable check used to be too strict as it lacked the property that
if an untyped const can be represented as a T, then it is assignable to T.

And we can now use that fixed check to add a missing check: in a return
statement, we now make sure that any of the returned elements are
assignable to what the signature tells us they should be.

### DIFF
--- a/_test/fun24.go
+++ b/_test/fun24.go
@@ -1,0 +1,10 @@
+package main
+
+func f(x int) (int, int) { return x, "foo" }
+
+func main() {
+	print("hello")
+}
+
+// Error:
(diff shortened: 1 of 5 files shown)


mpl:
For additional background: what prompted me to fix
`func (t *itype) assignableTo(o *itype) bool` is that it would fail if you try to use to check the return statement of `./_test/ret7.go`
And the go spec states that a return statement must be assignable.

mvertes (approved):
LGTM

PR #910, opened 2020-10-20
interp: added kind to log for typecheck

rsteube:
added `kind` to log
```
invalid operation: cannot take address of []github.com/rsteube/traefik-plugin-brotli/vendor/github.com/andybalholm/brotli.huffmanTree [kind: sliceExpr]
```


mvertes (approved):
Looks good to me

PR #912, opened 2020-10-21
interp: fix array size definition in case of forward declared constant

### LINKED ISSUE #911: array size definition

```go
package main

var a = [max]int{}

const max = 32

func main() {
    println(len(a))
}
```

Expected result:
```console
$ go run ./sample.go
32
```

Got:
```console
$ yaegi ./sample.go
./sample.go:3:10: undefined: max
```

mvertes:
In type.go, the symbol lookup on constant id was not performed. Handle
the ident kind explicitely to allow that.

Fixes #911.

### DIFF
--- a/_test/a44.go
+++ b/_test/a44.go
@@ -1,0 +1,12 @@
+package main
+
+var a = [max]int{}
+
+const max = 32
+
+func main() {
+	println(len(a))
+}
+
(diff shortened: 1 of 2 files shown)


nrwiersma (approved):
LGTM

PR #913, opened 2020-10-21
interp: added sliceExpr and starExpr

rsteube:
Type check was failing for expression such as: `&(*tree)[node:][0]`, as in

```go
tree, _ := huffmanTreePool.Get().(*[]huffmanTree)
...

initHuffmanTree(&(*tree)[node:][0], histogram[l], -1, int16(l))
```

see https://github.com/andybalholm/brotli/blob/c3da72aa01ed78f164593b9624fd91d25082d2d2/brotli_bit_stream.go#L469

### DIFF
--- a/.golangci.toml
+++ b/.golangci.toml
@@ -43,6 +43,10 @@
     "nestif",
     "exhaustive",
     "nlreturn",
+    "wrapcheck",
+    "exhaustivestruct",
+    "errorlint",
+    "tparallel",
   ]
 
 [issues]
--- a/.golangci.toml
+++ b/.golangci.toml
@@ -43,10 +43,6 @@
     "nestif",
     "exhaustive",
     "nlreturn",
-    "wrapcheck",
-    "exhaustivestruct",
-    "errorlint",
-    "tparallel",
   ]
 
 [issues]
--- a/_test/a44.go
+++ b/_test/a44.go
@@ -1,0 +1,12 @@
+package main
+
+var a = [max]int{}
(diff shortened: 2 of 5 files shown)


mvertes:
As a starting point, you first need to have a reproducible case, as simple as possible, which exposes the issue without the fix, and then demonstrates that the fix works. Simple scripts like the following are then added to the existing ones, for non regression testing (checking that your fix doesn't break something else).

So I suggest that you first open an issue, using the template, and provide a test case like this one, which corresponds to the issue:

```go
package main

import "fmt"
    
type T struct {
    A int
    B int
}
    
func main() {
    a := &[]T{
        {1, 2},
        {3, 4},
    }
    fmt.Println("a:", a)
    x := &(*a)[1:][0]
    fmt.Println("x:", x)
}
// Output:
// a: &[{1 2} {3 4}]
// x: &{3 4}
```

Then when you submit your PR, you can refer to the related issue with a line like `Fixes #xxx` in the description.

When testing with yaegi directly, to speed up the non regression tests, do `go test -short ./interp` rather than `go test ./...` or `make tests`, you will win time.

At this moment, the fix you propose is not sufficient, more investigation is required. I do not have the answer yet myself.

rsteube:
yeah expected that, just wanted this to be visible for discussion - i will open an issue

mvertes (approved):
The change you proposed was almost correct and I editted them directly. Thanks.

Looks good to me.

PR #930, opened 2020-10-29
interp: resolve type for untyped shift expressions

### LINKED ISSUE #927: type mismatch of shift on untyped expression

```go
package main

func main() {
    var max, a, b uint32
    max = a + (1 << b)
    println(max) 
}
```

Expected result:
```console
$ go run ./sample.go
1
```

Got:
```console
$ yaegi ./sample.go
./sample.go:4:6: panic
run: reflect: call of reflect.Value.Uint on int Value
goroutine 1 [running]:
runtime/debug.Stack(0x1, 0xc0002f4600, 0x40)
        /usr/lib/go/src/runtime/debug/stack.go:24 +0x9f
github.com/traefik/yaegi/interp.(*Interpreter).eval.func1(0xc000217c70)
        /home/marc/go/src/github.com/traefik/yaegi/interp/interp.go:480 +0xc5
panic(0xe01b60, 0xc00010ef00)
        /usr/lib/go/src/runtime/panic.go:969 +0x1b9
github.com/traefik/yaegi/interp.runCfg.func1(0xc0003780a0, 0xc00012cc00)
        /home/marc/go/src/github.com/traefik/yaegi/interp/run.go:124 +0x217
panic(0xe01b60, 0xc00010ef00)
        /usr/lib/go/src/runtime/panic.go:969 +0x1b9
reflect.Value.Uint(...)
        /usr/lib/go/src/reflect/value.go:1941
github.com/traefik/yaegi/interp.genValueUint.func2(0xc0003780a0, 0xdcdc60, 0xc00011fae4, 0x18a, 0x0)
        /home/marc/go/src/github.com/traefik/yaegi/interp/value.go:436 +0xfd
github.com/traefik/yaegi/interp.add.func9(0xc0003780a0, 0xc000434c60)
        /home/marc/go/src/github.com/traefik/yaegi/interp/op.go:95 +0x6b
github.com/traefik/yaegi/interp.runCfg(0xc00012cc00, 0xc0003780a0)
        /home/marc/go/src/github.com/traefik/yaegi/interp/run.go:130 +0x6e
github.com/traefik/yaegi/interp.(*Interpreter).run(0xc000372000, 0xc00012c400, 0xc000378000)
        /home/marc/go/src/github.com/traefik/yaegi/interp/run.go:108 +0x2af
github.com/traefik/yaegi/interp.(*Interpreter).eval(0xc000372000, 0xc00012f5c0, 0x56, 0x7fff3c07a4d1, 0x17, 0x0, 0x0, 0x0, 0x0, 0x0, ...)
        /home/marc/go/src/github.com/traefik/yaegi/interp/interp.go:572 +0x809
github.com/traefik/yaegi/interp.(*Interpreter).EvalPath(0xc000372000, 0x7fff3c07a4d1, 0x17, 0xc00012f500, 0x56, 0x0, 0x0, 0xc000123110)
        /home/marc/go/src/github.com/traefik/yaegi/interp/interp.go:394 +0xfa
main.runFile(0xc000372000, 0x7fff3c07a4d1, 0x17, 0xc000116240, 0x1)
        /home/marc/go/src/github.com/traefik/yaegi/cmd/yaegi/run.go:123 +0xae
main.run(0xc00000e090, 0x1, 0x1, 0x13, 0x13)
        /home/marc/go/src/github.com/traefik/yaegi/cmd/yaegi/run.go:89 +0x7e5
main.main()
        /home/marc/go/src/github.com/traefik/yaegi/cmd/yaegi/yaegi.go:144 +0x415
```

mvertes:
A non-constant shift expression can be untyped, requiring to apply a
type from inherited context. This change insures that such context is
propagated during CFG pre-order walk, to be used if necessary.
    
Fixes #927.

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -672,7 +672,6 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 			wireChild(n)
 			nilSym := interp.universe.sym[nilIdent]
 			c0, c1 := n.child[0], n.child[1]
-			rval0, rval1 := c0.rval, c1.rval
 
 			err = check.binaryExpr(n)
 			if err != nil {
@@ -680,10 +679,6 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 			}
 
 			switch n.action {
-			case aQuo:
-				// Restore original constant values as before check to allow automatic integer conversion.
-				// TODO(marc) To avoid this, constant values should be stored separately than rval.
-				c0.rval, c1.rval = rval0, rval1
 			case aRem:
 				n.typ = c0.typ
 			case aShl, aShr:
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -79,9 +79,26 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 					i--
 				}
 				dest := a.child[i]
-				n.typ = dest.typ
+				if dest.typ != nil && !isInterface(dest.typ) {
+					// Interface type are not propagated, and will be resolved at post-order.
+					n.typ = dest.typ
+				}
 			case binaryExpr, unaryExpr, parenExpr:
 				n.typ = n.anc.typ
+			}
+
+		case defineStmt:
+			// Determine type of variables initialized at declaration, so it can be propagated.
+			if n.nleft+n.nright == len(n.child) {
+				// No type was specified on the left hand side, it will resolved at post-order.
+				break
+			}
+			n.typ, err = nodeType(interp, sc, n.child[n.nleft])
+			if err != nil {
+				break
+			}
+			for i := 0; i < n.nleft; i++ {
+				n.child[i].typ = n.typ
 			}
 
 		case blockStmt:
@@ -465,7 +482,7 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 			var atyp *itype
 			if n.nleft+n.nright < len(n.child) {
 				if atyp, err = nodeType(interp, sc, n.child[n.nleft]); err != nil {
-					return
+					break
 				}
 			}
 
@@ -655,6 +672,7 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 			wireChild(n)
 			nilSym := interp.universe.sym[nilIdent]
 			c0, c1 := n.child[0], n.child[1]
+			rval0, rval1 := c0.rval, c1.rval
 
 			err = check.binaryExpr(n)
 			if err != nil {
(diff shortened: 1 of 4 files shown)


interp/cfg.go
```diff
@@ -860,7 +883,12 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 					n.gen = nop
 					n.findex = -1
 					n.typ = c0.typ
-					n.rval = c1.rval.Convert(c0.typ.rtype)
+					if c, ok := c1.rval.Interface().(constant.Value); ok {
+						i, _ := constant.Int64Val(constant.ToInt(c))
+						n.rval = reflect.ValueOf(i).Convert(c0.typ.rtype)
+					} else {
+						n.rval = c1.rval.Convert(c0.typ.rtype)
+					}
```
mpl, line 891: Does this have anything to do with the shift bug? If not, what is it about?

interp/cfg.go
```diff
@@ -644,7 +662,12 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 			}
 
 			switch n.action {
-			case aRem, aShl, aShr:
+			case aRem:
+				n.typ = c0.typ
+			case aShl, aShr:
+				if c0.typ.untyped {
+					break
+				}
```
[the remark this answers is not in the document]
mvertes: c0 is not necessarily a const here, It can be an untyped var i.e `1 << a` where `a` is a var. In that case we precisely want to use the context inherited type and not the one from child. The untyped criteria is the correct one.

interp/cfg.go
mvertes: this indeed relates to a different issue, but related also to shift, see interp_eval_test.go:78 for the corresponding test.

interp/cfg.go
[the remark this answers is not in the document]
mpl: But in your example (`1 << a`) , `c0` is not `a`, since `c0` is the 0th child of the expression, and `a` is at the end of the expression? no?

interp/cfg.go
mpl: ok. also, I'm not 100% sure, but I think instead of an if/else, we can have an if and break early here.

interp/cfg.go
```diff
@@ -66,6 +66,24 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 			return false
 		}
 		switch n.kind {
+		case binaryExpr, unaryExpr, parenExpr:
```
mpl, line 69: I understand why we need to catch all binary and paren expressions (and not just binary shift expressions), but I don't get why we need to catch unary expressions too. Can you give me an example?

mvertes: - (a * 2)

interp/cfg.go
mvertes: not worth for such small blocks

interp/cfg.go
mvertes: although this crashes right now, maybe another issue

mpl: I'm not sure what you meant, so out of curiosity I've tried:

```
package main

func main() {
    var max, a uint32
    max = 1 << - (a * 2)
    println(max) 
}
```

and `yaegi run` on that does not print anything. Maybe it's a different/another problem though, I haven't investigated further yet.

mpl: ok, let me know if you want to merge this PR as is and investigate that independently. Otherwise I'll wait for more updates on this PR.

mpl (approved):

PR #931, opened 2020-10-29
interp: on panic, look for node where offending exec originated from

### LINKED ISSUE #546: The line number of the code is incorrect while panic

```go
package main

import "fmt"

func testPanic2() {
	var m []int
	m[0] = 1
}

func testPanic() {
	testPanic2()
}

func main() {
	fmt.Println("hello")
	fmt.Println("hello")
	fmt.Println("hello")
	testPanic()
}
```

Expected result:
```console
$ go run ./sample.go
hello
hello
hello
7:6: panic
11:2: panic
18:2: panic
Exit:  reflect: slice index out of range
```

Got:
```console
$ yaegi ./sample.go
hello
hello
hello
6:6: panic
11:2: panic
15:2: panic
Exit:  reflect: slice index out of range
```

mpl:
The execution flow is such that a node can end up running several chained exec
funcs, some of which actually originate from other nodes. For example, in:

var m []int // L0
println("hello world") // L1
m[0] = 1 // L2

the offending code is on a node on line 2 (out of range error). However, since
the assignment to m is part of the execution flow of the variable m, we'll get
the panic when running all the chained exec funcs attached to the node for m on
line 0.

Which is why, when that happens, we need to look for the actual node (the one on
L2) where the offending instruction originates from, in order to
properly report the origin of the panic.

Fixes #546

### DIFF
--- a/interp/run.go
+++ b/interp/run.go
@@ -108,10 +108,53 @@ func (interp *Interpreter) run(n *node, cf *frame) {
 	runCfg(n.start, f)
 }
 
+// originalExecNode looks in the tree of nodes for the node which has exec,
+// aside from n, in order to know where n "inherited" that exec from.
+func originalExecNode(n *node, exec bltn) *node {
+	execAddr := reflect.ValueOf(exec).Pointer()
+	var originalNode *node
+	seen := make(map[int64]struct{})
+	root := n
+	for {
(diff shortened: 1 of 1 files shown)


mvertes (approved):
LGTM

PR #943, opened 2020-11-04
extract: support building/running yaegi with Go devel

### LINKED ISSUE #928: (limited) support for building w/ master
#### Proposal

it would be great if one could build `yægi` using the Go gc compiler's next version.
right now, when using the compiler from `master`, ad compiling `yægi`, one gets:

```
$> go get -v ./...
package github.com/traefik/yaegi/cmd/yaegi
	imports github.com/traefik/yaegi/stdlib/syscall: build constraints exclude all Go files in /home/binet/work/gonum/src/github.com/containous/yaegi/stdlib/syscall
package github.com/traefik/yaegi/cmd/yaegi
	imports github.com/traefik/yaegi/stdlib/unsafe: build constraints exclude all Go files in /home/binet/work/gonum/src/github.com/containous/yaegi/stdlib/unsafe
```

the standard library doesn't insert that many new symbols from one release cycle to the next, so it seems to me it would be relatively cheap to just add the build constraints for `n+1` whenever the Go release `n` is cut.

added benefit: earlier bug reports w/ the next Go release that `yægi` will have to support.

mpl:
Fixes #928


extract/extract.go
```diff
@@ -436,10 +436,12 @@ func GetMinor(part string) string {
...
 }
 
+const defaultMinorVersion = 15
+
 func genBuildTags() (string, error) {
 	version := runtime.Version()
-	if version == "devel" {
-		return "", nil
+	if strings.HasPrefix(version, "devel") {
+		return fmt.Sprintf("1.%d", defaultMinorVersion), nil
```
mpl, line 444: actually we might not want to do that, I'll think about it again

mvertes: I think that this change is not necessary. As maintainers, we want to run `extract` only for stable Go versions. By removing the constraint on next version, it is still possible to re-build yaegi with next and have a working stdlib.

May be add a comment in the Makefile to warn against the necessity to run `go generate` on with an `devel` version of Go compiler (but not forbid to do so).

mpl: I know that the extract/ lib can be used by other projects, and not just for extracting the stdlib for yaegi, so it might make sense to allow "devel", but at the moment genBuildTags is actually only called in the case of the stdlib, for which we do not want to allow "devel" at the moment. So I think it's legitimate to return an error for now. And it makes the failure explicit when someone tries to run "make generate" with it. We can always change it back to be more lenient later, if we ever call genBuildTags for other things. WDYT?

mvertes: ok

mvertes (commented):
LGTM

mvertes (approved):

PR #956, opened 2020-11-10
interp: support calling goto from sub-scope

### LINKED ISSUE #953: conditional goto
The following program `sample.go` triggers an unexpected result:

```go
package main

func main() {
    if true {
        goto here
    }
here:
    println("ok")
}
```

Expected result:
```console
$ go run ./sample.go
ok
```

Got:
```console
$ yaegi ./sample.go
```

mvertes:
As opposed to other symbols, goto labels must be searched in included
scopes, not upper ones. Implement scope.lookdown to perform this,
to allow calls to goto to be embedded in included scopes where label
is defined.

Fixes #953.

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -194,7 +194,7 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 
 		case breakStmt, continueStmt, gotoStmt:
 			if len(n.child) > 0 {
-				// Handle labeled statements
+				// Handle labeled statements.
 				label := n.child[0].ident
 				if sym, _, ok := sc.lookup(label); ok {
 					if sym.kind != labelSym {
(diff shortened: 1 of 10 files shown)


mvertes:
Detection of wrong goto is missing, and will be done in a later change.

interp/cfg.go
```diff
@@ -211,25 +211,22 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 
 		case labeledStmt:
 			label := n.child[0].ident
-			if sym, _, ok := sc.lookup(label); ok {
-				if sym.kind != labelSym {
-					err = n.child[0].cfgErrorf("label %s not defined", label)
-					break
-				}
```
mpl, line 218: It seems to me that this check (that the symbol we found is indeed a label kind), is orthogonal to the actual change (of lookup becoming lookdown). Do you confirm that you removed this check because it was actually never needed?

interp/scope.go
```diff
@@ -130,6 +132,20 @@ func (s *scope) lookup(ident string) (*symbol, int, bool) {
 	return nil, 0, false
 }
 
+// lookdown searches for a symbol in the current scope and included ones, recursively.
+// It returns the first found symbol and true, or nil and false.
```
mpl, line 136: if the returned symbol is always nil when not found, then its "nilness" can always be used as the sentinel in the caller, and the boolean is not actually needed. Not a big deal though.

mpl: hmm, but I suppose have a (val,ok) form allows the caller to do "if _,ok := sc.lookdown", which I imagine may allow some memory optimizations when the compiler knows the first returned element is not needed and can be skipped. so maybe it's always better to have a comma-ok form...

mvertes: I am not entirely sure that a symbol with a `nil` value is always invalid, and also I'd like to keep API consistency with symetrical `lookup` method.

interp/cfg.go
mvertes: This check can be removed.

I just found  that goto labels can not be stored in scope symbols, but need a serparate namespace. In any case the current implementation is flawed, and does not support collisions between goto labels and other symbols, as shown by the following:

```go
package main

func main() {
    here := 0
    println("before goto")
here:
    if here > 0 {
        return
    }
    here++
    println("ok", here)
    goto here
}
```

Expected:
```console
$ go run ./sample.go
before goto
ok 1
```

Got:
```console
$ yaegi ./sample.go
./sample.go:6:1 label here not defined
```

So I propose to merge this PR as is, as it is better than current situtation, and open a new issue with the above example regarding label/symbol collision, which will require a deeper change in the way the labels are stored.

mpl (approved):

PR #958, opened 2020-11-13
interp: fix type check on function signature

### LINKED ISSUE #957: Regression in v0.9.4: undefined selector Time
I found a regression in Yaegi, it appeared in v0.9.4, it affects the plugin `plugin-simplecache`:

```
error: /tmp/pilot-gop211396919/src/github.com/traefik/plugin-simplecache/cache.go:12:2: import \"github.com/pquerna/cachecontrol\"
error: /tmp/pilot-gop211396919/src/github.com/traefik/plugin-simplecache/vendor/github.com/pquerna/cachecontrol/api.go:21:2: import \"github.com/pquerna/cachecontrol/cacheobject\"
error: /tmp/pilot-gop211396919/src/github.com/traefik/plugin-simplecache/vendor/github.com/pquerna/cachecontrol/cacheobject/object.go:255:32: undefined selector Time"
```

https://github.com/pquerna/cachecontrol/blob/master/cacheobject/object.go

https://github.com/traefik/yaegi/compare/v0.9.3...v0.9.4

related to https://github.com/traefik/plugin-simplecache/issues/12

mvertes:
Perform function declaration type check from the upper level scope (the scope where the
function is declared), to avoid possible collisions of local variables with package names.

Fixes #957.

### DIFF
--- a/_test/time14.go
+++ b/_test/time14.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"fmt"
+	"time"
+)
+
+var t time.Time
+
+func f() time.Time {
+	time := t
+	return time
+}
+
+func main() {
+	fmt.Println(f())
+}
+
+// Output:
+// 0001-01-01 00:00:00 +0000 UTC
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -1381,7 +1381,7 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 			n.val = sc.def
 			for i, c := range n.child {
 				var typ *itype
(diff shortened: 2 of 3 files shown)


mpl:
> Perform function declaration type check from the global scope, to avoid
> possible collisions of local variables with package names.
> 
> Fixes #957.

Is this commit message above still correct? Because in this one you say we use the global scope, but then in the latest commit you say "use upperLevel instead of global". So which is it? When for each child of the return statement we do
```
typ, err = nodeType(interp, sc.upperLevel(), returnSig.child[1].fieldType(i))
```
are we using the global scope or not?

mvertes:
I clarified the commit message.

mpl (approved):

PR #966, opened 2020-11-25
interp: support more type assertion cases

### LINKED ISSUE #955: interface implementation assertion fails
Similar, to #919, but with an interface defined in the interpreter scope:

```go
package main

import (
	"fmt"
	"log"
)

type MyWriter interface {
	Write(p []byte) (n int, err error)
}

type TestStruct struct{}

func (t TestStruct) Write(p []byte) (n int, err error) {
	fmt.Println(len(p))
	return len(p), nil
}

func main() {
	var t interface{}
	t = TestStruct{}
	tw, ok := t.(MyWriter)
	if !ok {
		println("nope")
	} else {
	    tw.Write(nil)
        }
}
```

Expected result:
```console
$ go run ./sample.go
0
```

Got:
```console
$ yaegi ./sample.go
nope
```

mpl:
Fixes #955

### DIFF
--- a/_test/assert0.go
+++ b/_test/assert0.go
@@ -1,0 +1,58 @@
+package main
+
+import (
+	"fmt"
+	"time"
+)
+
+type MyWriter interface {
+	Write(p []byte) (i int, err error)
+}
+
+type TestStruct struct{}
+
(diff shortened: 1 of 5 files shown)


mvertes (approved):
LGTM

PR #988, opened 2020-12-14
interp: add support for binary composite slice

### LINKED ISSUE #987: automatic type in literal composite

```go
package main

import (
    "fmt"
    "net/url"
)

func main() {
    body := url.Values{
        "Action": {"none"},
    }   
    fmt.Println(body)
}   
```

Expected result:
```console
$ go run ./sample.go
map[Action:[none]]
```

Got:
```console
$ yaegi ./sample.go
./sample.go:14:13: undefined type
```

mvertes:
Fixes #987.

### DIFF
--- a/interp/run.go
+++ b/interp/run.go
@@ -2213,6 +2213,51 @@ func compositeBinMap(n *node) {
 	}
 }
 
+func compositeBinSlice(n *node) {
+	value := valueGenerator(n, n.findex)
+	next := getExec(n.tnext)
+	child := n.child
+	if n.nleft == 1 {
+		child = n.child[1:]
+	}
+
+	values := make([]func(*frame) reflect.Value, len(child))
+	index := make([]int, len(child))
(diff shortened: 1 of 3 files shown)


interp/run.go
```diff
@@ -2213,6 +2213,51 @@ func compositeBinMap(n *node) {
...
+	value := valueGenerator(n, n.findex)
+	next := getExec(n.tnext)
+	child := n.child
+	if n.nleft == 1 {
+		child = n.child[1:]
+	}
+
+	values := make([]func(*frame) reflect.Value, len(child))
+	index := make([]int, len(child))
+	rtype := n.typ.rtype.Elem()
```
mpl, line 2226: Maybe we also need the part here to handle the cases for a slice of pointers, like in doCompositeBinStruct:

	if n.typ.cat == ptrT || n.typ.cat == aliasT {
		rtype = n.typ.val.rtype
	}

?

mpl (approved):

Issue #995, opened 2021-01-04
Constant definition loop with protoreflect


katallaxie:
The following library [protoreflect](https://github.com/protocolbuffers/protobuf-go/blob/master/reflect/protoreflect/methods.go) triggers a panic.

It creates the following panic.

```bash
google.golang.org/protobuf/reflect/protoreflect/methods.go:18:2: constant definition loop
```

mvertes:
also related to #260

katallaxie:
@mvertes With #1001 the error has changed.

```bash
vendor/google.golang.org/protobuf/reflect/protoreflect/value_unsafe.go:48:9: panic
vendor/google.golang.org/protobuf/reflect/protoreflect/value_unsafe.go:42:16: panic
```

It seems that the `EnumNumber` https://github.com/protocolbuffers/protobuf-go/blob/160c7477e0e899d5072bb25635f46053df619fbf/reflect/protoreflect/value_unsafe.go#L42 is not imported.

mvertes:
Sorry for being so late.

The fix here is to use the pure go version of the package. This is done by using yaegi tags with the build tags as in https://github.com/protocolbuffers/protobuf-go/blob/160c7477e0e899d5072bb25635f46053df619fbf/reflect/protoreflect/value_pure.go#L5.

I tested the following with yaegi-v0.12.0 and it works (at least the import):

sample.go:
```go
package main

// yaegi:tags purego appengine

import ( 
    _ "google.golang.org/protobuf/reflect/protoreflect"
)

func main() {
    println("ok")
}
```

and then:
```console
$ yaegi ./sample.go
ok
```

Issue #260, opened 2019-07-24
Panic on import "google.golang.org/grpc"


zliang-min:
Steps:
1. run `yaegi`
2. run `import "google.golang.org/grpc"`

It then will panic:

```console
> import "google.golang.org/grpc"
/Users/gimi/workspace/go/src/golang.org/x/net/context/go17.go:15:15: panic
panic: reflect.Set: value of type context.Context is not assignable to type uint32 [recovered]
        panic: reflect.Set: value of type context.Context is not assignable to type uint32

goroutine 1 [running]:
github.com/containous/yaegi/interp.runCfg.func1(0xc0000eb590, 0xc0003b3d00)
        /Users/gimi/workspace/go/pkg/mod/github.com/containous/yaegi@v0.0.1/interp/run.go:107 +0x1e9
panic(0x17fcf20, 0xc00038d200)
        /usr/local/Cellar/go/1.12.7/libexec/src/runtime/panic.go:522 +0x1b5
reflect.Value.assignTo(0x18644c0, 0xc00038d1f0, 0x94, 0x191e2a9, 0xb, 0x17fd1a0, 0x0, 0xc000364f00, 0x1, 0x1)
        /usr/local/Cellar/go/1.12.7/libexec/src/reflect/value.go:2339 +0x437
reflect.Value.Set(0x17fd1a0, 0xc00031c8c0, 0x18a, 0x18644c0, 0xc00038d1f0, 0x94)
        /usr/local/Cellar/go/1.12.7/libexec/src/reflect/value.go:1473 +0xa8
github.com/containous/yaegi/interp.callBin.func5(0xc0000eb590, 0xc0003660f0)
        /Users/gimi/workspace/go/pkg/mod/github.com/containous/yaegi@v0.0.1/interp/run.go:838 +0x269
github.com/containous/yaegi/interp.runCfg(0xc0003b3d00, 0xc0000eb590)
        /Users/gimi/workspace/go/pkg/mod/github.com/containous/yaegi@v0.0.1/interp/run.go:112 +0x70
github.com/containous/yaegi/interp.(*Interpreter).run(0xc0000c4540, 0xc0003b3200, 0x0)
        /Users/gimi/workspace/go/pkg/mod/github.com/containous/yaegi@v0.0.1/interp/run.go:93 +0x126
[...]

PR #1001, opened 2021-01-12
interp: parse circular interface definitions

### LINKED ISSUE #999: recursive interface with aliased type
`The` following program `sample.go` triggers an unexpected error:

```go
package main

import "fmt"

type I1 interface{ A }

type A = I2
    
type I2 interface{ F() I1 }
    
func main() {
    var i I1
    fmt.Println(i)
}
```

Expected result:
```console
$ go run ./sample.go
<nil>
```

Got:
```console
$ yaegi ./sample.go
./sample.go:5:6: constant definition loop
```

This is a simplified test case from what we see in #260 and #995.

mvertes:
An undefined type detection function has been added to better diagnose
incomplete type definitions. Implicit type names in interface or struct
declarations are now better handled. The incomplete status is not
fowarded to aliased type declarations to handle circular definitions.

Fixes #999 and #995. Improves #260 (goes farther, but still fails).

### DIFF
--- a/interp/gta.go
+++ b/interp/gta.go
@@ -319,7 +319,56 @@ func (interp *Interpreter) gtaRetry(nodes []*node, importPath string) error {
 	}
 
 	if len(revisit) > 0 {
-		return revisit[0].cfgErrorf("constant definition loop")
+		for _, n := range revisit {
+			if n.kind == typeSpec {
+				if err := definedType(n.typ); err != nil {
+					return err
+				}
+			}
+		}
+		switch revisit[0].kind {
+		case typeSpec:
+			return revisit[0].cfgErrorf("incomplete type definition: %s", revisit[0].typ.name)
+		default:
+			return revisit[0].cfgErrorf("constant definition loop")
+		}
+	}
+	return nil
+}
+
+func definedType(typ *itype) error {
+	if !typ.incomplete {
+		return nil
+	}
+	switch typ.cat {
+	case interfaceT, structT:
+		for _, f := range typ.field {
+			if err := definedType(f.typ); err != nil {
+				return err
+			}
+		}
+	case funcT:
+		for _, t := range typ.arg {
+			if err := definedType(t); err != nil {
+				return err
+			}
+		}
+		for _, t := range typ.ret {
+			if err := definedType(t); err != nil {
+				return err
+			}
+		}
+	case mapT:
+		if err := definedType(typ.key); err != nil {
+			return err
+		}
+		fallthrough
+	case aliasT, arrayT, chanT, chanSendT, chanRecvT, ptrT, variadicT:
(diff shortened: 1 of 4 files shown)


interp/type.go
```diff
@@ -885,7 +888,13 @@ func isComplete(t *itype, visited map[string]bool) bool {
 		visited[name] = true
 	}
 	switch t.cat {
-	case aliasT, arrayT, chanT, chanRecvT, chanSendT, ptrT:
+	case aliasT:
```
mpl, line 891: Sorry, because github is stupid, I have to comment here.

I believe L884-L886 can be simplified as:

```
if visited[name] {
   return true
}
```

interp/type.go
```diff
@@ -885,7 +888,13 @@ func isComplete(t *itype, visited map[string]bool) bool {
...
+		if t.val != nil && t.val.incomplete && t.val.cat != nilT {
```
mpl, line 892: If the aliased type is already complete, we also want to return true, right?
```suggestion
		if t.val != nil && t.val.cat != nilT {
```

interp/type.go
```diff
@@ -899,6 +908,8 @@ func isComplete(t *itype, visited map[string]bool) bool {
 	case interfaceT, structT:
 		complete := true
 		for _, f := range t.field {
+			// Fiedld implicit type names must be marked as visited, to break false circles.
+			visited[f.typ.path+"/"+f.typ.name] = true
 			complete = complete && isComplete(f.typ, visited)
```
mpl, line 913:
```suggestion
			complete &= isComplete(f.typ, visited)
```

interp/gta.go
```diff
@@ -319,7 +319,56 @@ func (interp *Interpreter) gtaRetry(nodes []*node, importPath string) error {
...
+		for _, n := range revisit {
+			if n.kind == typeSpec {
+				if err := definedType(n.typ); err != nil {
+					return err
+				}
+			}
+		}
+		switch revisit[0].kind {
+		case typeSpec:
+			return revisit[0].cfgErrorf("incomplete type definition: %s", revisit[0].typ.name)
```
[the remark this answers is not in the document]
mvertes: The type could still be incomplete even if defined, if for example an array size is still undefined (constant undefined). But in that case, we should directly display "constant undefined" instead, so yes, you are right, something has to be changed here.

interp/type.go
mvertes: No! It's not equivalent.You are replacing a logical `and` where right side is not evaluated if left is false, with an arithmetic bitwise `and` (not boolean), where the operand is called unconditionally.

mpl: indeed

mpl (approved):

PR #1008, opened 2021-01-14
interp: fix detection of type recursivity

### LINKED ISSUE #1007: Panic on complex embedded structures

```go
package main

import "github.com/traefik/yaegi/interp"

const src = `package foo

type TypeA struct {
	B TypeB
}

type TypeB struct {
	C1 *TypeC
	C2 *TypeC
}

type TypeC struct {
	Val string
	D *TypeD
	D2 *TypeD
}

type TypeD struct {
	Name        string
}

func build() *TypeA {
	return &TypeA{
		B: TypeB{
			C2: &TypeC{Val: "22"},
		},
	}
}
func Bar(s string) string { 
	a := build()
	return s + "-" + a.B.C2.Val
}
`

func main() {

	_, err := i.Eval(src)
	if err != nil {
		panic(err)
	}

	v, err := i.Eval("foo.Bar")
	if err != nil {
		panic(err)
	}

	bar := v.Interface().(func(string) string)

	r := bar("Kung")
	println(r)
}
```

Expected result:
```console
$ go run ./sample.go
// output
Kung-22
```

Got:
```console
$ go run ./sample.go
// output
25:9: panic
30:7: panic
panic: reflect.Set: value of type struct { C1 *struct { Val string; D *struct { Name string }; D2 *struct { Name string } }; C2 *struct { Val string; D *struct { Name string }; D2 *struct { Name string } } } is not assignable to type struct { C1 *interface {}; C2 *interface {} } [recovered]
        panic: reflect.Set: value of type struct { C1 *struct { Val string; D *struct { Name string }; D2 *struct { Name string } }; C2 *struct { Val string; D *struct { Name string }; D2 *struct { Name string } } } is not assignable to type struct { C1 *interface {}; C2 *interface {} } [recovered]
        panic: reflect.Set: value of type struct { C1 *struct { Val string; D *struct { Name string }; D2 *struct { Name string } }; C2 *struct { Val string; D *struct { Name string }; D2 *struct { Name string } } } is not assignable to type struct { C1 *interface {}; C2 *interface {} }
```

mvertes:
If a struct contains several fields of the same temporary incomplete
type, it could be detected incorrectly as a recursive struct. Pass
a copy of defined types map to avoid this issue.

Fixes #1007.

### DIFF
--- a/interp/type.go
+++ b/interp/type.go
@@ -1551,7 +1551,7 @@ func hasRecursiveStruct(t *itype, defined map[string]*itype) bool {
 		defined[typ.path+"/"+typ.name] = typ
 
 		for _, f := range typ.field {
-			if hasRecursiveStruct(f.typ, defined) {
+			if hasRecursiveStruct(f.typ, copyDefined(defined)) {
 				return true
 			}
 		}
--- a/_test/issue-1007.go
+++ b/_test/issue-1007.go
@@ -1,0 +1,40 @@
+package main
+
+type TypeA struct {
+	B TypeB
+}
+
+type TypeB struct {
+	C1 *TypeC
+	C2 *TypeC
+}
+
+type TypeC struct {
+	Val string
+	D   *TypeD
+	D2  *TypeD
+}
+
+type TypeD struct {
+	Name string
+}
+
+func build() *TypeA {
+	return &TypeA{
+		B: TypeB{
+			C2: &TypeC{Val: "22"},
+		},
+	}
+}
+
+func Bar(s string) string {
+	a := build()
+	return s + "-" + a.B.C2.Val
+}
+
+func main() {
+	println(Bar("test"))
+}
+
+// Output:
+// test-22


interp/type.go
```diff
@@ -1540,7 +1540,7 @@ func hasRecursiveStruct(t *itype, defined map[string]*itype) bool {
 		defined[typ.path+"/"+typ.name] = typ
 
 		for _, f := range typ.field {
-			if hasRecursiveStruct(f.typ, defined) {
+			if hasRecursiveStruct(f.typ, copyDefined(defined)) {
```
mpl, line 1543: I'm wondering: is it not a problem in general that copyDefined is only copying pointers of itypes instead of making deepcopies of them? is that intended?

mpl: ah nevermind, I see we're never writing the stored itypes, it's all readonly, so it should ok.

mvertes: HasRecursive only tracks if a type refers to itself, using `defined` as an ephemeral map to mark them as visited (as in the type dependencies tree). There is no sense to deepcopy types here. But it is important that 2 sibling sub-types remain isolated, thus the `copyDefined` function.

mpl (approved):

PR #1020, opened 2021-01-28
fix: untyped check

### LINKED ISSUE #1000: assign imported variable get type mismatch

```go
package main

import (
	"fmt"
	"reflect"

)

const expression = `test_val == 11"`

func main() {
	i.Use(stdlib.Symbols)
	i.Use(interp.Exports{
		"mytest": {
			"test_val": reflect.ValueOf(int64(11)),
		},
	})
	init := `
import "mytest"

var test_val = mytest.test_val
`
	_, err := i.Eval(init)
	if err != nil {
		panic(err)
	}
	v, err := i.Eval(expression)
	if err != nil {
		panic(err)
	}
	fmt.Println(v.Interface().(bool))
}
```

Expected result:
```console
true
```

Got:
```console
4:5: panic
panic: reflect.Set: value of type int64 is not assignable to type int
```

nrwiersma:
When checking for untyped values, we can be sure at this stage that they must be a const value or already untyped. Checking for type string equality is no longer a good measure.

Fixes #1000

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -2570,11 +2570,7 @@ func isValueUntyped(v reflect.Value) bool {
 	if v.CanSet() {
 		return false
 	}
-	t := v.Type()
-	if t.Implements(constVal) {
-		return true
-	}
-	return t.String() == t.Kind().String()
+	return v.Type().Implements(constVal)
 }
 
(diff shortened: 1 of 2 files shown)


mvertes (approved):
LGTM

Issue #886, opened 2020-10-05
Yaegi panics when using a func in a template FuncMap


nrwiersma:
The following program `sample.go` triggers a panic:

```go
package main

import (
	"bytes"
	"html/template"
)

var str = `{{ stringOr .Data "test" }}`

func main() {
	tmpl, err := template.New("test").
		Funcs(template.FuncMap{
			"stringOr": stringOr,
		}).
		Parse(str)
	if err != nil {
		println(err.Error())
		return
	}

	data := map[string]interface{}{"Data": ""}

	var buf bytes.Buffer
	if err  := tmpl.Execute(&buf, data); err != nil {
		println(err.Error())
		return
	}
	println(buf.String())
}

func stringOr(v, def string) string {
	if v == "" {
		return def
	}
	return v
}

```

Expected result:
```console
$ go run ./sample.go
test
```

Got:
```console
$ yaegi ./sample.go
./_test/test.go:11:15: panic
value for stringOr not a function
goroutine 1 [running]:
runtime/debug.Stack(0x1, 0xc0001fec00, 0x40)
	/usr/local/Cellar/go/1.15.2/libexec/src/runtime/debug/stack.go:24 +0x9f
github.com/traefik/yaegi/interp.(*Interpreter).eval.func1(0xc000113d08)
	/Users/nrwiersma/Go/src/github.com/traefik/yaegi/interp/interp.go:451 +0xc5
panic(0x19bbec0, 0xc00049a0d0)
	/usr/local/Cellar/go/1.15.2/libexec/src/runtime/panic.go:969 +0x175
github.com/traefik/yaegi/interp.runCfg.func1(0xc0002466e0, 0xc0001c9e00)
	/Users/nrwiersma/Go/src/github.com/traefik/yaegi/interp/run.go:124 +0x217
panic(0x19bbec0, 0xc00049a0d0)
	/usr/local/Cellar/go/1.15.2/libexec/src/runtime/panic.go:969 +0x175
text/template.addValueFuncs(0xc000494630, 0xc000494690)
	/usr/local/Cellar/go/1.15.2/libexec/src/text/template/funcs.go:92 +0x4c9
text/template.(*Template).Funcs(0xc0002e70c0, 0xc000494690, 0x0)
	/usr/local/Cellar/go/1.15.2/libexec/src/text/template/template.go:174 +0x9b
html/template.(*Template).Funcs(0xc000494660, 0xc000494690, 0x0)
[...]

PR #1021, opened 2021-01-28
fix: handle function references in composite bin map

nrwiersma:
When passing a function reference as an interface in a composite binary map, the case should be handled to not take the value of the the node.

Related to #886


mvertes (approved):
LGTM

PR #1023, opened 2021-01-29
chore: use GitHub Actions.

ldez:
- use GitHub Actions instead of TravisCI
- updates golangci-lint to v1.36.0 and applies my rules


mvertes (commented):
LGTM

PR #1033, opened 2021-02-08
add support for sandboxes (packages with plugins or extensions)

althink:
Adds possibility to create a sandboxed dir where plugins/extensions can use/import each other.

For example:
`i := interp.New(interp.Options{
	Sandboxes: map[string]string{
		"ext": "extensions",
	},
})`

From now on it's possible to add packages with logic to "./extensions" dir and to import from a script:
`import "ext/plugin1"`

additionally any extension can import/use code from the other just by importing "ext/plugin2"

Thanks to this structure it's possible to invoke "go mod init ext" in extension directory and we can write extensions with full IDE support.

### DIFF
--- a/interp/src.go
+++ b/interp/src.go
@@ -33,6 +33,8 @@ func (interp *Interpreter) importSrc(rPath, importPath string, skipTest bool) (s
 			rPath = "."
 		}
 		dir = filepath.Join(filepath.Dir(interp.name), rPath, importPath)
+	} else if sandboxPath, ok := findSandboxPath(interp.sandboxes, importPath); ok {
+		dir = sandboxPath
 	} else if dir, rPath, err = interp.pkgDir(interp.context.GOPATH, rPath, importPath); err != nil {
 		// Try again, assuming a root dir at the source location.
 		if rPath, err = interp.rootFromSourceLocation(); err != nil {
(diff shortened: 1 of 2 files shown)


interp/src.go
```diff
@@ -307,3 +309,13 @@ func isPathRelative(s string) bool {
 	p := "." + string(filepath.Separator)
 	return strings.HasPrefix(s, p) || strings.HasPrefix(s, "."+p)
 }
+
+// findSandboxPath checks if given import path belongs to some sandbox.
+func findSandboxPath(sandboxes map[string]string, importPath string) (string, bool) {
+	for k, v := range sandboxes {
+		if strings.HasPrefix(importPath, k+"/") {
```
mewmew, line 316: Potentially it could be worth the clean the import path before prefix comparison (i.e. `path.Clean(importPath)`). This may help mitigate path traversal vulnerabilities should users try to escape the file system sandbox root directory.

For instance, given:

```go
 i := interp.New(interp.Options{ Sandboxes: map[string]string{ "ext": "extensions", }, })
```

The importPath `"k/../foo"` would pass the prefix check and result in the path `"extensions/../foo"` -> `"foo"` being returned.

I may be missing some details, just wanted to post this for consideration while the PR is active.

Cheers,
Robin

octalide:
Any interest in getting this moving again? I'm building a project that would benefit greatly from a use case like this one.

mvertes:
Thank you for this proposal, and sorry for the delay to respond.

It seems to me that the intent here is to implement something close to Go modules, but without having to parse a `go.mod` file. Is it right ?

Before moving further, I would like to see an actual code sample using this feature. You could add it in the `example` directory.

Thanks

dennwc:
I implemented something similar for one of my projects. But instead of changing the interpreter, I made a custom FS implementation that only allowed accessing specific directories - very similar to how `chroot` would work. Having a custom FS also allows it to completely ignore symlinks, etc to make it more secure. The code is longer than in this PR, but allows more control, I think.

mvertes:
I understand the need of isolating and restricting the access to the filesystem in a sandboxing approach. I think that it could be implemented on top of what already exists in yaegi to virtualise the restricted stdlib functions. I also wonder how the stdlib package `io/fs` could be of help.

As for implementing the support of Go modules, it should be addressed separately.

So I'm closing this PR, and we can restart the conversation in the discussions if necessary.

PR #1034, opened 2021-02-12
fix: use portable path manipulation for choosing package names

whereswaldon:
This *should* fix the generated names of packages on windows, which currently take the form `C:\kdfjslfj\jkfsjldkfjsf\sjdkfldjf` (this was reported to me by someone attempting to generate yaegi symbols for one of my projects). That being said, I don't have a windows machine to test on.


JackMordaunt:
> 
> 
> This _should_ fix the generated names of packages on windows, which currently take the form `C:\kdfjslfj\jkfsjldkfjsf\sjdkfldjf` (this was reported to me by someone attempting to generate yaegi symbols for one of my projects). That being said, I don't have a windows machine to test on.

Can confirm. Fixes the issue on windows.

mvertes (approved):
Looks good to me

PR #1038, opened 2021-02-16
feat: update stdlib mapping for go1.16

ldez:
- drop stdlib for go1.14
- generate stdlib for go1.16
-  update CI configuration.


mvertes (approved):
LGTM

PR #1039, opened 2021-02-18
Use fmt.Fprintln instead of fmt.Println in runCfg

### LINKED ISSUE #1030: Send existing stacktrace from paniced Eval to the configured interp.Options.Stderr
#### Proposal

At the moment (0.9.11) the stacktrace when an Eval panics goes only to the apps `fmt.Println()` and not to the configured `interp.Options.Stderr`. This makes it hard to capture the stack trace when you wish to surface that info to somewhere other than the apps stdout.

If yaegi also printed the stacktrace to the configured `interp.Options.Stderr` it would make it possible for me to consume it, and logically it makes sense to me that the stacktrace would appear in the scripts configured Stderr.

I believe this is the line that prints it - https://github.com/traefik/yaegi/blob/master/interp/run.go#L182 

#### Background

I `recover()` from Eval panics and wish to show the existing stacktrace (that only seems to go to the apps stdout) to the user in the GUI.

#### Workarounds

I could not find one, but maybe I missed another way of getting the stacktrace from a panic'ed eval?

I am also compiling for JS/WASM where os.Pipe() is not available, so I cannot simply replace and consume main apps Stdout/err to get the stacktrace when the Eval panics.

tigerwill90:
As mentioned in #1030, when an Eval panic, it is print with `fmt.Println()` and not to the configured `interp.Options.Stderr`. According to https://github.com/traefik/yaegi/blob/master/interp/interp.go#L210, it should be removed in future version so I'm not sure if this pull request is necessary. However, it could fix the issue in the meanwhile.

Fixes #1030.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -208,7 +208,7 @@ type Panic struct {
 }
 
 // TODO: Capture interpreter stack frames also and remove
-// fmt.Println(n.cfgErrorf("panic")) in runCfg.
+// fmt.Fprintln(n.interp.stderr, oNode.cfgErrorf("panic")) in runCfg.
 
 func (e Panic) Error() string { return fmt.Sprint(e.Value) }
 
--- a/interp/run.go
+++ b/interp/run.go
@@ -180,7 +180,7 @@ func runCfg(n *node, f *frame) {
 			if oNode == nil {
 				oNode = n
 			}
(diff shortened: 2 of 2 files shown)


mvertes (commented):
Looks good to me. Thank you for your contribution.

ldez (approved):

PR #1044, opened 2021-02-25
interp: fix a memory management issue causing wrong closure context

### LINKED ISSUE #1035: closure captures incorrect value
The following program `yaegi_closure_problem.go` gives unexpected output:

```go
package main

import "fmt"

type Foo struct {
        GetName func() string
        GetNum  func() int
}

func main() {
        foos := []Foo{}

        for i, txt := range []string{"zero", "one", "two", "three"} {
                nameForClosure := fmt.Sprintf("%d %s", i, txt)
                iForClosure := i
                foos = append(foos, Foo{
                        GetName: func() string { return nameForClosure },
                        GetNum:  func() int { return iForClosure },
                })
        }

        for i := range foos {
                fmt.Println(
                        foos[i].GetName(),
                        foos[i].GetNum(),
                )
        }
}
```

Expected result:
```console
$ go run ./yaegi_closure_problem.go
0 zero 0
1 one 1
2 two 2
3 three 3
```

Got:
```console
$ yaegi ./yaegi_closure_problem.go
3 three 3
3 three 3
3 three 3
3 three 3
```

mvertes:
The first change forces a variable definition to reallocate a
new memory slot to avoid corrupting a previously defined one in
a loop block.

The second change ensures that the frame clone operations obtains
a copy of the original data slice, to preserve the original context
set in a loop.

Fixes #1035.

### DIFF
--- a/_test/closure10.go
+++ b/_test/closure10.go
@@ -1,0 +1,19 @@
+package main
+
+func main() {
+	foos := []func(){}
+
+	for i := 0; i < 3; i++ {
+		a, b := i, i
+		_ = b
+		foos = append(foos, func() { println(i, a) })
+	}
+	foos[0]()
+	foos[1]()
+	foos[2]()
+}
+
+// Output:
+// 3 0
+// 3 1
+// 3 2
--- a/_test/closure10.go
+++ b/_test/closure10.go
@@ -5,8 +5,7 @@ func main() {
 
 	for i := 0; i < 3; i++ {
 		a, b := i, i
-		_ = b
-		foos = append(foos, func() { println(i, a) })
+		foos = append(foos, func() { println(i, a, b) })
 	}
 	foos[0]()
 	foos[1]()
@@ -14,6 +13,6 @@ func main() {
 }
 
 // Output:
-// 3 0
-// 3 1
-// 3 2
+// 3 0 0
+// 3 1 1
+// 3 2 2
--- a/interp/run.go
+++ b/interp/run.go
@@ -924,19 +924,8 @@ func genFunctionWrapper(n *node) func(*frame) reflect.Value {
 		}
 	}
 	funcType := n.typ.TypeOf()
-	notInDefer := n.anc != nil && n.anc.anc != nil && n.anc.anc.kind != deferStmt
 
 	return func(f *frame) reflect.Value {
-		if n.findex >= 0 && notInDefer {
-			// Generate a function wrapper on the node value in the frame
-			// obtained by getFunc, rather than the static node value,
-			// to preserve the right closuse context.
-			if v := getFrame(f, n.level).data[n.findex]; v.IsValid() {
-				if nod, ok := v.Interface().(*node); ok {
-					n = nod
-				}
-			}
-		}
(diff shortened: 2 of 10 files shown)


interp/run.go
```diff
@@ -642,6 +642,15 @@ func assign(n *node) {
 				d(f).SetMapIndex(i(f), s(f))
 				return next
 			}
+		case n.kind == defineStmt:
```
mpl, line 645: Why are we handling only the n.nleft == 1 case (L634) ? i.e. if I slightly change the new test to:

```
package main

func main() {
	foos := []func(){}

	for i := 0; i < 3; i++ {
		a, b := i, i
		_ = a
		foos = append(foos, func() { println(i, b) })
	}
	foos[0]()
	foos[1]()
	foos[2]()
}

// Output:
// 3 0
// 3 1
// 3 2
```

then it still does not work.

mpl (commented):
I've noticed two problems:

one is that we still do not handle multiple assignments on the same line (see comment in code).

the other is that I tested the case from the issue, i.e.:

```
package main

import "fmt"

type Foo struct {
        GetName func() string
        GetNum  func() int
}

func main() {
        foos := []Foo{}

        for i, txt := range []string{"zero", "one", "two", "three"} {
                nameForClosure := fmt.Sprintf("%d %s", i, txt)
                iForClosure := i
                foos = append(foos, Foo{
                        GetName: func() string { return nameForClosure },
                        GetNum:  func() int { return iForClosure },
                })
        }

        for i := range foos {
                fmt.Println(
                        foos[i].GetName(),
                        foos[i].GetNum(),
                )
        }
}
```
and it looks like it is still broken (but I haven't investigated why).
So this PR is a progress, but I don't think it fully fixes #1035 yet at the moment.

interp/run.go
mvertes: fixed now, and test case added

_test/closure10.go
```diff
@@ -0,0 +1,19 @@
+package main
+
+func main() {
+	foos := []func(){}
+
+	for i := 0; i < 3; i++ {
+		a, b := i, i
+		_ = b
+		foos = append(foos, func() { println(i, a) })
```
mpl, line 9: You could do:
```suggestion
		foos = append(foos, func() { println(i, a, b) })
```
to prove that b actually has the right value

interp/run.go
```diff
@@ -657,45 +658,70 @@ func assign(n *node) {
...
+		index[i] = n.child[i].findex
+		level[i] = n.child[i].level
+
+	}
 
-		// To handle swap in multi-assign:
-		// evaluate and copy all values in assign right hand side into temporary
-		// then evaluate assign left hand side and copy temporary into it
+	if n.kind == defineStmt {
+		// Handle a multiple var declararation / assgin. It cannot be a swap.
```
mpl, line 685:
```suggestion
		// Handle a multiple var declararation / assign. It cannot be a swap.
```

mpl (commented):
It looks like the case from https://github.com/traefik/yaegi/issues/1035 still does not work. Should I review carefully the code, or are you still working on it?

mpl (approved):

PR #1066, opened 2021-03-25
interp: allow redeclaration of imports

firelizzard18:
This PR adds an interpreter option, `AllowRedeclaration`. If this option is set, `(*Interpreter).Eval` will allow package imports to be redeclared. That is, no error will be raised and the package symbol will be overwritten.

I would like to use Yaegi to power a Go notebook (VSCode extension), somewhat like Jupyter. A notebook can have multiple Go 'cells' which can be evaluated (using Yaegi). As much as is possible, evaluating cells should be idempotent - that is, evaluating a cell multiple times should have the same effect as evaluating it once, ideally. Cells that are not idempotent can degrade the user experience.

Specifically, Go files tend to declare all imports in a single block. In a notebook, I'd put all imports in a single block, in their own cell. When I decide I need to import an additional package, I want to add that import to the existing cell and evaluate it. Without this MR, reevaluating that block usually causes an error.

### DIFF
--- a/interp/gta.go
+++ b/interp/gta.go
@@ -209,8 +209,11 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
 					// imports in different source files of the same package. Therefore, we suffix
 					// the key with the basename of the source file.
 					name = filepath.Join(name, baseName)
-					if _, exists := sc.sym[name]; interp.allowRedecl || !exists {
+					if sym, exists := sc.sym[name]; !exists {
 						sc.sym[name] = &symbol{kind: pkgSym, typ: &itype{cat: binPkgT, path: ipath, scope: sc}}
+						break
+					} else if sym.kind == pkgSym && sym.typ.cat == srcPkgT && sym.typ.path == ipath {
+						// ignore re-import of identical package
 						break
 					}
 
@@ -233,8 +236,11 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
 						name = pkgName
 					}
 					name = filepath.Join(name, baseName)
-					if _, exists := sc.sym[name]; interp.allowRedecl || !exists {
+					if sym, exists := sc.sym[name]; !exists {
 						sc.sym[name] = &symbol{kind: pkgSym, typ: &itype{cat: srcPkgT, path: ipath, scope: sc}}
+						break
+					} else if sym.kind == pkgSym && sym.typ.cat == srcPkgT && sym.typ.path == ipath {
+						// ignore re-import of identical package
 						break
 					}
 
--- a/interp/gta.go
+++ b/interp/gta.go
@@ -209,7 +209,7 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
 					// imports in different source files of the same package. Therefore, we suffix
 					// the key with the basename of the source file.
 					name = filepath.Join(name, baseName)
-					if _, exists := sc.sym[name]; !exists {
+					if _, exists := sc.sym[name]; interp.allowRedecl || !exists {
 						sc.sym[name] = &symbol{kind: pkgSym, typ: &itype{cat: binPkgT, path: ipath, scope: sc}}
 						break
 					}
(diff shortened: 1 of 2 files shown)


interp/gta.go
```diff
@@ -233,7 +233,7 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
 						name = pkgName
 					}
 					name = filepath.Join(name, baseName)
-					if _, exists := sc.sym[name]; !exists {
+					if _, exists := sc.sym[name]; interp.allowRedecl || !exists {
```
mvertes, line 236: I think we can always tolerate a  re-import of an identical package in yaegi, as packages to parse are de-duplicated in any case. Adding an `AllowRedecl` option to interp is useless to me, there is no sense nor value to be too strict here, beyond your use case. I suggest to remove the definition of `AllowRedecl` option and instead verifying that if the symbol already exists, it matches the existing one:
```
if sym, exists := sc.sym[name]; exists && sim.kind == pkgSym && sym.typ.cat == srcPkgT && sym.typ.path == ipath {
    break
}

File: interp/gta.go [hunk 2]
```diff
@@ -209,7 +209,7 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
 					// imports in different source files of the same package. Therefore, we suffix
 					// the key with the basename of the source file.
 					name = filepath.Join(name, baseName)
-					if _, exists := sc.sym[name]; !exists {
+					if _, exists := sc.sym[name]; interp.allowRedecl || !exists {
```
mvertes, line 212: see my comment for line 236 below, and do the same

interp/interp.go
```diff
@@ -124,13 +124,14 @@ type opt struct {
...
-	stdout   io.Writer     // standard output
-	stderr   io.Writer     // standard error
+	dotCmd      string
+	noRun       bool          // compile, but do not run
+	fastChan    bool          // disable cancellable chan operations
+	context     build.Context // build context: GOPATH, build constraints
+	stdin       io.Reader     // standard input
+	stdout      io.Writer     // standard output
+	stderr      io.Writer     // standard error
+	allowRedecl bool          // allow redeclaration
```
mvertes, line 134: no need for allowRedecl, see my comment above gta.go:236

interp/interp.go
```diff
@@ -244,6 +245,9 @@ type Options struct {
 	// They default to os.Stding, os.Stdout and os.Stderr respectively.
 	Stdin          io.Reader
 	Stdout, Stderr io.Writer
+
```
mvertes, line 248: No need for this declaration.

interp/interp.go
```diff
@@ -261,6 +265,8 @@ func New(options Options) *Interpreter {
 		hooks:    &hooks{},
 	}
 
+	i.opt.allowRedecl = options.AllowRedeclaration
```
mvertes, line 268: no need for this declaration

mvertes (changes requested):
Thanks for your submission. I think it can be simplified a bit, as I think that an additional option is not really required.

interp/gta.go
firelizzard18: 👍

interp/interp.go
firelizzard18: 👍

interp/interp.go
firelizzard18: 👍

interp/interp.go
firelizzard18: 👍

interp/gta.go
firelizzard18: 👍

firelizzard18:
@mvertes I applied your suggestion:

```go
if sym, exists := sc.sym[name]; !exists {
	sc.sym[name] = &symbol{kind: pkgSym, typ: &itype{cat: binPkgT, path: ipath, scope: sc}}
	break
} else if sym.kind == pkgSym && sym.typ.cat == srcPkgT && sym.typ.path == ipath {
	// ignore re-import of identical package
	break
}
```

mvertes (approved):
LGTM

PR #1070, opened 2021-03-30
Allow colon-equal to override variable type

firelizzard18:
If I execute the following:

```
I := interp.New(interp.Options{})
I.Eval(`x := 1`)
I.Eval(`x := "foobar"`)
```

I expect the second declaration to override the first. `var x string` will override the previous type, and redeclaring a type, function, const, etc will override it, but the `:=` operator will not.

Currently, the result is: `reflect.Set: value of type string is not assignable to type int`

### DIFF
--- a/interp/gta.go
+++ b/interp/gta.go
@@ -80,9 +80,7 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
 				if typ.isBinMethod {
 					typ = &itype{cat: valueT, rtype: typ.methodCallType(), isBinMethod: true, scope: sc}
 				}
-				if sc.sym[dest.ident] == nil || sc.sym[dest.ident].typ.incomplete {
-					sc.sym[dest.ident] = &symbol{kind: varSym, global: true, index: sc.add(typ), typ: typ, rval: val, node: n}
-				}
+				sc.sym[dest.ident] = &symbol{kind: varSym, global: true, index: sc.add(typ), typ: typ, rval: val, node: n}
 				if n.anc.kind == constDecl {
 					sc.sym[dest.ident].kind = constSym
 					if childPos(n) == len(n.anc.child)-1 {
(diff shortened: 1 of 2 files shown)


mvertes:
Redefining symbols (a const or a var in the same scope) with a different type is potentially problematic and should be avoided  in the general case. This is not an idempotent operation, like re-importing the same package. We should rather raise an error (if feasible) rather than generalising it. Do you really need that feature ?

firelizzard18:
To be clear, this change does not allow a variable to be redeclared multiple times within a given evaluated string. Go's AST parser still rejects that.

I'm using Yaegi as an interactive environment programming environment (REPL, notebooks, etc) - I expect the execution context in an interactive environment to be mutable. Granted, most such environments are built around dynamic languages, which don't have the same issues.

> Redefining symbols (a const or a var in the same scope) with a different type is potentially problematic and should be avoided in the general case.

I can see how this would be an issue when using Yaegi to drive a scripting system. However, in a user-driven interactive system, it goes against my expectations.

The specific case where I've run into this is something like the following:

https://user-images.githubusercontent.com/879055/113346034-ebc47e80-92f8-11eb-897c-98d39771fdb7.mov

Personally, the inconsistency of Yaegi's behavior is enough reason: I cannot change the type of a variable with `x := <value>` but I can with `var x <type>`

I don't know why my screencap clobbered the error message. It was:

```
1:28: panic
reflect.Set: value of type *struct { Baz string "json:\"baz\"" } is not assignable to type *struct { Baz string }
```

mvertes:
Ok, thanks. So let's give it a try. And by the way, impressive video of the netbook use.

mvertes (approved):
LGTM

firelizzard18:
> impressive video of the netbook use

Thanks! VSCode's Notebook API is not yet finalized, so I can't publish the extension yet, but when I do I hope people will find it useful. The [README](https://gitlab.com/ethan.reesor/vscode-notebooks/go-kernel/-/blob/master/README.md) declares, "Powered by Yaegi", so the extension's users will know where credit is due for the core functionality.

PR #1079, opened 2021-04-14
Add an option to suppress output from assignments

firelizzard18:
This MR adds an option, `interp.Options.SuppressAssignmentResult`, which causes assignments to output `interp.EmptyResult`.


mvertes:
The current state is not consistent. It would be better to always disable result output for assignment rather than adding an option, for consistency and simplicity. Could you modify your PR in that direction ?
Thanks

firelizzard18:
@mvertes This is ready for review. Now, all variable declarations and assignments return the special value `interp.EmptyResult`.

PR #1080, opened 2021-04-14
interp: implement unsafe.Offsetof

### LINKED ISSUE #1062: Feature: add Offsetof to unsafe
#### Proposal
I want to add unsafe.Offsetof, can you help me? @nrwiersma

```go
func offsetof(s interface{}, f string) uintptr {
	typ := reflect.ValueOf(s).Type()
	sf, ok := typ.FieldByName(f)
	if !ok {
		panic(fmt.Sprintf("structure %s not contain field: \"%s\"", typ, f))
	}
	return sf.Offset
}
```

This code can get structure field offset, but I don't know how to add it to the yaegi interpreter,
for use like
```go
unsafe.Offsetof(StructureA{}.FieldA)
```

I can only add a simple api but with two parameter, offset(s interface{}, f string) to "stdlib/unsafe".
#### Background
unsafe.Offsetof

#### Workarounds
unsafe.Offsetof

mvertes:
Offsetof returns the offset of a field in a struct. It is computed
during parsing at CFG, due to the constraint of operating on a
struct selector expression.

With this function, the support of 'unsafe' package is now
complete in yaegi.

Fixes #1062.

### DIFF
--- a/_test/unsafe5.go
+++ b/_test/unsafe5.go
@@ -12,11 +12,13 @@ type S struct {
 }
 
 func main() {
-	size := unsafe.Sizeof(S{})
-	align := unsafe.Alignof(S{})
+	x := S{}
+	size := unsafe.Sizeof(x)
+	align := unsafe.Alignof(x.Y)
+	offset := unsafe.Offsetof(x.Z)
 
-	fmt.Println(size, align)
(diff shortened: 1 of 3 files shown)


nrwiersma (approved):
LGTM

PR #1081, opened 2021-04-14
Add option to produce array of results for blocks

firelizzard18:
This PR adds an option, `interp.Options.BlockStatementAsArray`, which configures the interpreter to return an array of results for each statement of a block, instead of just the last statement.


mvertes:
Thank you for your proposal, but please, wait a little so we can settle the previous ones. It's not a go yet for #1078 and #1079, we need a bit of a design review first before jumping on coding, and I still have concerns. I will provide more details soon.

firelizzard18:
FWIW, all three PRs are independent features. They don't build on each other.

PR #1096, opened 2021-04-28
interp: fix type string extraction for wrapped constants

### LINKED ISSUE #1089: Imported stdlib string constant cannot be concatenated to a string literal

```go
package main

import (
        "fmt"
        "time"
)

func main() {
        fmt.Println(`"` + time.RFC3339Nano + `"`)
}
```

Expected result:
```console
$ go run ./sample.go
"2006-01-02T15:04:05.999999999Z07:00"
```

Got:
```console
$ yaegi ./sample.go
run: ./sample.go:9:14: invalid operation: mismatched types string and
```

Found in the wild here: https://github.com/Azure/go-autorest/blob/6184171f42979bea0ac05eb62163f6eb5d9e0d45/autorest/date/time.go#L26

I managed to get it to work on my local system by making the following change to extract.go (and regenerating stdlib/go1_16_time.go):

```diff
diff --git a/extract/extract.go b/extract/extract.go
index f6b59b3..f8844fe 100644
--- a/extract/extract.go
+++ b/extract/extract.go
@@ -299,9 +299,6 @@ func fixConst(name string, val constant.Value, imports map[string]bool) string {
                str string
        )
        switch val.Kind() {
-       case constant.String:
-               tok = "STRING"
-               str = val.ExactString()
        case constant.Int:
                tok = "INT"
                str = val.ExactString()
```

mvertes:
Fixes #1089.

### DIFF
--- a/interp/type.go
+++ b/interp/type.go
@@ -1188,8 +1188,7 @@ func (t *itype) id() (res string) {
 }
 
 func constTypeString(t reflect.Type) (s string) {
-	e := reflect.New(t).Elem()
-	cv, ok := e.Interface().(constant.Value)
+	cv, ok := reflect.New(t).Elem().Interface().(constant.Value)
 	if !ok {
 		return
 	}
--- a/interp/type.go
+++ b/interp/type.go
@@ -1172,6 +1172,10 @@ func (t *itype) id() (res string) {
 		}
 		res += "}"
 	case valueT:
+		if isConstantValue(t.rtype) {
+			res = constTypeString(t.rtype)
+			break
+		}
 		res = ""
 		if t.rtype.PkgPath() != "" {
 			res += t.rtype.PkgPath() + "."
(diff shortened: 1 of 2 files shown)


interp/type.go
```diff
@@ -1183,6 +1187,26 @@ func (t *itype) id() (res string) {
 	return res
 }
 
+func constTypeString(t reflect.Type) (s string) {
+	cv, ok := reflect.New(t).Elem().Interface().(constant.Value)
+	if !ok {
+		return
+	}
```
mpl, line 1194: we're returning the empty string here, but in the caller we do not check whether res is empty, so I think this is going to end up being silently ignored, won't it?
So are you sure you don't want to panic instead?

mvertes: Returning an empty string if unresolved is correct for use in `id()`. We do not want to panic here, as it would avoid us to return a proper error during parse.

mpl (approved):

PR #1102, opened 2021-04-30
interp: fix package name extraction from import path

### LINKED ISSUE #1095: Custom built interpreter (with extra binary packages compiled in) detects wrong package name for extracted symbol
Following the instructions here: https://github.com/traefik/yaegi/issues/856#issuecomment-695766266
I've compiled an interpreter to add the binary package "github.com/dgrijalva/jwt-go" to the interpreter.

However, this fails on the following test program:

```go
package main

import (
        "github.com/dgrijalva/jwt-go"
        "fmt"
)

func main() {
        claims := jwt.MapClaims{}
        fmt.Println(claims)
}
```

Expected result:
```console
$ go run ./sample.go
map[]
```

Got:
```console
$ customyaegi ./sample.go
*interp.cfgError sample.go:9:12: undefined type
```

I believe the cause is that jwt-go is detected as having package name "go" by the interpreter, rather than the package name "jwt" that it actually has.

I managed to get this to work on my local system by having the extract command pass through the package name. Not sure if this is the right solution. Excerpts of this patch below. The full patch is much too long to fit in here, due to all the references that have to be adjusted..

```diff
diff --git a/interp/interp.go b/interp/interp.go
index 4d0c593..264081f 100644
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -112,8 +112,15 @@ func (f *frame) clone(fork bool) *frame {
        return nf
 }

+type ExportSymbols map[string]reflect.Value
+
+type Package struct {
+       PkgName string
+       ExportSymbols
+}
+
 // Exports stores the map of binary packages per package path.
-type Exports map[string]map[string]reflect.Value
+type Exports map[string]Package
diff --git a/interp/gta.go b/interp/gta.go
index a1db163..5f4369e 100644
--- a/interp/gta.go
+++ b/interp/gta.go
@@ -200,7 +200,7 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
                                        }
                                default: // import symbols in package namespace
                                        if name == "" {
-                                               name = identifier.FindString(ipath)
+                                               name = interp.binPkg[ipath].PkgName
                                        }
                                        // imports of a same package are all mapped in the same scope, so we cannot just
                                        // map them by their names, otherwise we could have collisions from same-name
```

mvertes:
The heuristic to generate a package name identifier was incorrect. Now for binary packages, the package identifier is obtained by a symbol, generated by extract, which contains the string argument of package statement in source file. This should ensure an always correct default package identifier. 

Fixes #1095.


bailsman:
Would this pick "go" for e.g. https://github.com/pelletier/go-toml ?

mvertes:
@bailsman , good point. With this patch, the package name would be `go`, which is still unfortunate.

We may have to exclude Go keywords(`go`, `if`, `for`, ...) and maybe pre-reserved identifiers such as `int`, `bool`, etc. The specification leaves the choice to the implementer, as far. as I understand, but we 'd better follow Go tool (still need to retrieve the exact details).

bailsman:
The package name being part of the path may just be a convention that is not enforced by the compiler.
The following package is named "gobbledy" but lives at github.com/bailsman/arbitrary-package

```go
package main

import (
        "github.com/bailsman/arbitrary-package"
        "fmt"
)

func main() {
        fmt.Println(gobbledy.Gook)
}
```

The go compiler accepts it:

```console
$ go run ./sample.go
Package names can be arbitrary and don't need to be in the path.
```

Locally, I've modified the extract command to make the extracted symbols a struct with a package name (string) field and an exports (map[string]reflect.Value) field.

mvertes:
You are right in that the default package name must be obtained from the `package` statement in the source file, and not from the import path.

The package name should also be made available by `extract`, as you mention, but it could be stored in a symbol with a special non identifier key (i.e. `.package`), instead of changing the data structure. For example:
```go
import (
    "github.com/dgrijalva/jwt-go"
    "reflect"
)

func init() {
    Symbols["github.com/dgrijalva/jwt-go"] = map[string]reflect.Value{
        // default package name identifier
       ".package":  reflect.ValueOf("jwt")

        // function, constant and variable definitions
        "DecodeSegment":                         reflect.ValueOf(jwt.DecodeSegment),
        "EncodeSegment":                         reflect.ValueOf(jwt.EncodeSegment),
[....]
```

Processing of interpreted packages names has also to be checked (note for myself).

interp/gta.go
```diff
@@ -187,11 +187,14 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
...
-			if interp.binPkg[ipath] != nil {
+			if pkg := interp.binPkg[ipath]; pkg != nil {
 				switch name {
 				case "_": // no import of symbols
 				case ".": // import symbols in current scope
-					for n, v := range interp.binPkg[ipath] {
+					for n, v := range pkg {
+						if n[0] == '.' {
+							continue
+						}
```
mpl, line 197: I do not understand that part. what is it for?

interp/interp.go
```diff
@@ -751,8 +758,8 @@ func ignoreScannerError(e *scanner.Error, s string) bool {
 // rename packages.
 func (interp *Interpreter) ImportUsed() {
 	sc := interp.universe
-	for k := range interp.binPkg {
-		name := key2name(k)
+	for k, pkg := range interp.binPkg {
+		name := key2name(pkg[".name"].String())
```
mpl, line 762: Is it a leftover from the previous commit before we decided to store the package name in the key, instead of in .name?
If not, I do not understand what that is for.

interp/gta.go
mvertes: It's a leftover from previous attempt, and should be removed. Thanks for noticing it.

interp/interp.go
mvertes: Yes, a leftover from previous attempt. It should be removed.

mpl (approved):

PR #1103, opened 2021-05-01
interp: sandbox to preserve type of os.Stdin os.Stdout and os.Stderr

### LINKED ISSUE #1092: Cannot assign os.Stdout to type *os.File

```go
package main

import (
        "fmt"
        "os"
        "reflect"
)

func main() {
        var x (*os.File)
        x = os.Stdout
        fmt.Println(reflect.TypeOf(x))
}
```

Expected result:
```console
$ go run ./sample.go
*os.File
```

Got:
```console
$ yaegi ./sample.go
run: ./sample.go:11:6: cannot use type io.Writer as type *os.File in assignment
```

Found in the real world in: https://github.com/Azure/go-autorest/blob/820a319d2214dc91d083022ae5ec50070175314c/logger/logger.go#L229

mvertes:
Use YAEGI_SPECIAL_STDIO env boolean to overwrite os.Stdxxx by
non file descriptors io.Writer / io.Reader interfaces. It is set
to true when testing to allow redirection to byte buffers.

The default behaviour is now to preserve the original concrete type
when sandboxing stdio, which maintains compatibility.

Fixes #1092.

### DIFF
--- a/interp/interp.go
+++ b/interp/interp.go
@@ -128,13 +128,14 @@ type opt struct {
 	cfgDot bool // display CFG graph (debug)
 	// dotCmd is the command to process the dot graph produced when astDot and/or
 	// cfgDot is enabled. It defaults to 'dot -Tdot -o <filename>.dot'.
-	dotCmd   string
-	noRun    bool          // compile, but do not run
-	fastChan bool          // disable cancellable chan operations
-	context  build.Context // build context: GOPATH, build constraints
-	stdin    io.Reader     // standard input
-	stdout   io.Writer     // standard output
-	stderr   io.Writer     // standard error
+	dotCmd       string
+	noRun        bool          // compile, but do not run
+	fastChan     bool          // disable cancellable chan operations
+	context      build.Context // build context: GOPATH, build constraints
+	specialStdio bool          // Allows os.Stdin, os.Stdout, os.Stderr to not be file descriptors
+	stdin        io.Reader     // standard input
+	stdout       io.Writer     // standard output
+	stderr       io.Writer     // standard error
 }
 
 // Interpreter contains global resources and state.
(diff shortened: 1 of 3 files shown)


interp/interp.go
```diff
@@ -721,9 +726,23 @@ func fixStdio(interp *Interpreter) {
...
 	if p = interp.binPkg["os"]; p != nil {
-		p["Stdin"] = reflect.ValueOf(&stdin).Elem()
-		p["Stdout"] = reflect.ValueOf(&stdout).Elem()
-		p["Stderr"] = reflect.ValueOf(&stderr).Elem()
+		if interp.specialStdio {
+			// Inherit streams from interpreter even if they do not have a file descriptor.
+			p["Stdin"] = reflect.ValueOf(&stdin).Elem()
+			p["Stdout"] = reflect.ValueOf(&stdout).Elem()
+			p["Stderr"] = reflect.ValueOf(&stderr).Elem()
+		} else {
```
mpl, line 734: To be honest, I now think that maybe it was a mistake that we allowed that in the past, and we should stop doing it.
But it's not urgent and we can do it later, and the PR fixes the current problem, so ok for now.

I'll approve but I'll let you merge yourself, to be sure you are certain you want to keep going this way.

mpl (approved):
When you say:

> The default behaviour is now to preserve the original concrete type
> when sandboxing stdio, which maintains compatibility.

you mean that we now have compatibility with the os pkg, right? But we're actually breaking compatibility with previous versions of the interpreter since we are changing the default behaviour. :-)

mvertes:
I agree. Let's see if allowing non file descriptor based stdio causes more problems in the near future. If yes we just disable it (and rework tests relying on it). It can be done later, merging for now.

PR #1109, opened 2021-05-08
interp: add example to show use of Self

kortschak:
This example clarifies how `interp.Self` can be accessed.

Please take a look.


mvertes (approved):
LGTM

PR #1112, opened 2021-05-12
interp: fix special range on string

### LINKED ISSUE #1088: range over multibyte character string yields wrong rune position
The following program `sample.go` gives an unexpected result:

```go
package main

import "fmt"

func main() {
    for i, ch := range "日本語" {
        fmt.Printf("%#U starts at byte position %d\n", ch, i)
    }
}
```

Expected result:
```console
$ go run ./sample.go
U+65E5 '日' starts at byte position 0
U+672C '本' starts at byte position 3
U+8A9E '語' starts at byte position 6
```

Got:
```console
$ yaegi ./sample.go
U+65E5 '日' starts at byte position 0
U+672C '本' starts at byte position 1
U+8A9E '語' starts at byte position 2
```

mvertes:
In this range variant "for k, v := range aString", k must
be the byte position of the rune in the byte array, rather than
the index of the rune in the rune array.

Fixes #1088.

### DIFF
--- a/interp/run.go
+++ b/interp/run.go
@@ -2629,6 +2629,8 @@ func _range(n *node) {
 		an = n.child[2]
 		index1 := n.child[1].findex // array value location in frame
 		if isString(an.typ.TypeOf()) {
+			// Special variant of "range" for string, where the index indicates the byte position
+			// of the rune in the string, rather than the index of the rune in array.
 			stringType := reflect.TypeOf("")
 			value = genValueAs(an, rat) // range on string iterates over runes
 			n.exec = func(f *frame) bltn {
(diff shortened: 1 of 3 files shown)


interp/run.go
```diff
@@ -2615,31 +2615,50 @@ var rat = reflect.ValueOf((*[]rune)(nil)).Type().Elem() // runes array type
 func _range(n *node) {
 	index0 := n.child[0].findex // array index location in frame
 	index2 := index0 - 1        // shallow array for range, always just behind index0
+	index3 := index2 - 1        // additional location to stop string char position
```
mpl, line 2618:
```suggestion
	index3 := index2 - 1        // additional location to store string char position
```
?

mpl (approved):

PR #1120, opened 2021-05-19
interp: fix method lookup on pointers to binary types

### LINKED ISSUE #1083: undefined field or method: WriteTo
The code I am using to run the interpreter is -> https://github.com/ShadowNight573/Discord-Bot-Test/blob/main/bot/commands/debug/eval.go#L32

The code I am interpreting with it is -> 
```
package test

import "github.com/bwmarrin/discordgo"

func test() string {
    var session *discordgo.Session
    fmt.Printf("%T", session)
    return "Session type"
}
```

The error I am getting is -> image

The code the error is coming from is -> https://github.com/gorilla/websocket/blob/master/conn_write.go#L13

---

This issue came up after #1077 was solved. Not sure how to solve this issue either, the packages work fine outside of the interpreter as far as I know.

mvertes:
This case was missing in the selector expression processing.

Fixes #1083.

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -1490,28 +1490,28 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 						n.typ = &itype{cat: valueT, rtype: field.Type}
 						n.val = field.Index
 						n.gen = getPtrIndexSeq
-					} else {
-						err = n.cfgErrorf("undefined field or method: %s", n.child[1].ident)
+						break
 					}
+					err = n.cfgErrorf("undefined field or method: %s", n.child[1].ident)
 				case n.typ.rtype.Kind() == reflect.Struct:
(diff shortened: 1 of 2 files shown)


interp/cfg.go
```diff
@@ -1488,28 +1488,28 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
...
-						} else {
-							err = n.cfgErrorf("undefined field or method: %s", n.child[1].ident)
-						}
+						break
 					}
+					fallthrough
 				default:
+					// method lookup failed on type, now lookup on pointer to type
+					pt := reflect.PtrTo(n.typ.rtype)
+					if m2, ok2 := pt.MethodByName(n.child[1].ident); ok2 {
```
mpl, line 1505: I haven't seen in the code above whether we make sure that n.typ.rtype is not nil before we try that. But I'll assume it is safe to do it even if n.typ.rtype is nil...

mpl (approved):

PR #1140, opened 2021-06-13
interp: handle binary array literals

firelizzard18:
This is a small change that allows use of composite array literals, such as:

```go
type Vec3 [3]float32

var foo = []Vec3{
	{1, 0, 0},
	{6, 0, 0},
	{6, 2, 0},
	{2, 2, 0},
	{1, 1, 0},
}
```


mvertes (approved):
LGTM

PR #1143, opened 2021-06-15
Add a newline after src in wrapInMain

### LINKED ISSUE #1142: Interpreter.Eval fails if the last line of the input is a comment
If `Interpreter.Eval` is called with `s` and `s` is not terminated with a newline and the last line of `s` contains `//`, evaluation fails because the closing brace of the generated main function is commented out.

```go
package main

import "github.com/traefik/yaegi/interp"

func main() {
	_, err := i.Eval(`
		println("Hello world")
		// foo bar`)
	if err != nil {
		panic(err)
	}
}
```

Expected result:
```console
$ go run ./sample.go
Hello world
```

Got:
```console
$ go run ./sample.go
panic: _.go:3:14: expected '}', found 'EOF'

goroutine 1 [running]:
main.main()
        .../sample.go:11 +0xbb
exit status 2
```

firelizzard18:
Fixes #1142

### DIFF
--- a/interp/ast.go
+++ b/interp/ast.go
@@ -357,7 +357,7 @@ func ignoreError(err error, src string) bool {
 }
 
 func wrapInMain(src string) string {
-	return fmt.Sprintf("package main; func main() {%s}", src)
+	return fmt.Sprintf("package main; func main() {%s\n}", src)
 }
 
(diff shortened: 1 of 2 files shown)


mvertes (approved):
LGTM

PR #1146, opened 2021-06-16
interp: fix assign of function values with binary methods

### LINKED ISSUE #1145: assign a function value with a binary method

```go
package main

import "sync"

type F func()

func main() {
    var wg sync.WaitGroup
    var f F = wg.Done
    println(f != nil)
}   
```

Expected result:
```console
$ go run ./sample.go
true
```

Got:
```console
$ yaegi ./sample.go
run: /Users/marc/sample.go:9:12: cannot use type  as type main.F in assignment
```

mvertes:
In typecheck.go, detect binary methods so we know when to skip the receiver as first parameter when checking function signatures. The signature check is not yet performed, we just avoid a false error.

In cfg.go, take care to label types with isBinMethod field to true whenever a binary method is resolved.

Also, do not attempt to wrap node in functions if the node value is already a binary function.

Fixes #1145.

### DIFF
--- a/interp/value.go
+++ b/interp/value.go
@@ -157,8 +157,8 @@ func genValueAsFunctionWrapper(n *node) func(*frame) reflect.Value {
 		if v.IsNil() {
 			return reflect.New(typ).Elem()
 		}
-		vn := v.Interface().(*node)
-		if vn.rval.IsValid() && vn.rval.Type().Kind() == reflect.Func {
+		vn, ok := v.Interface().(*node)
+		if ok && vn.rval.IsValid() && vn.rval.Type().Kind() == reflect.Func {
 			// The node value is already a callable func, no need to wrap it.
 			return vn.rval
 		}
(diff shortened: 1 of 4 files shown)


interp/value.go
```diff
@@ -157,7 +157,12 @@ func genValueAsFunctionWrapper(n *node) func(*frame) reflect.Value {
 		if v.IsNil() {
 			return reflect.New(typ).Elem()
 		}
-		return genFunctionWrapper(v.Interface().(*node))(f)
+		vn := v.Interface().(*node)
+		if vn.rval.IsValid() && vn.rval.Type().Kind() == reflect.Func {
```
mpl, line 161: Is it useless to instead do:
```suggestion
		vn, ok := v.Interface().(*node)
		if ok && vn.rval.IsValid() && vn.rval.Type().Kind() == reflect.Func {
```
?

mpl (approved):

PR #1152, opened 2021-06-17
Fix handling of ImportUsed packages

### LINKED ISSUE #1151: undefined type error for composite literal expression when type is included via Used
```go
func TestIssue1151(t *testing.T) {
	type pkgType []int

	i.Use(interp.Exports{
		"pkg/pkg": map[string]reflect.Value{
			"Type": reflect.ValueOf((*pkgType)(nil)),
		},
	})
	i.ImportUsed()

	runTests(t, i, []testCase{
		{src: "pkg.Type{1}", res: "[1]"},
	})
}
```

Expected result: test succeeds.

Got:
```console
$ go test -run=TestIssue1151 ./interp/
--- FAIL: TestIssue1151 (0.00s)
    --- FAIL: TestIssue1151/#00 (0.00s)
        interp_eval_test.go:784: got an error: 1:28: undefined type
FAIL
FAIL    github.com/traefik/yaegi/interp 0.004s
FAIL
```

firelizzard18:
Fixes #1151

If I add a package with `Use` and import it with `ImportUsed`, the package is added to the universe scope as `<pkg>`. If I import with `Eval`, the package is added as `<pkg>/_.go`. However, `(*node).isType` (in cfg.go) only checks for `<pkg>/_.go`. Thus, packages imported with `ImportUsed` can be inaccessible.

This MR updates `(*node).isType` to fall back to `<pkg>` if `<pkg>/_.go` does not exist.

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -2186,7 +2186,10 @@ func (n *node) isType(sc *scope) bool {
 		suffixedPkg := filepath.Join(pkg, baseName)
 		sym, _, ok := sc.lookup(suffixedPkg)
 		if !ok {
-			return false
+			sym, _, ok = sc.lookup(pkg)
+			if !ok {
+				return false
+			}
 		}
 		if sym.kind != pkgSym {
 			return false
--- a/interp/interp_eval_test.go
+++ b/interp/interp_eval_test.go
@@ -1511,3 +1511,22 @@ func TestIssue1142(t *testing.T) {
 		{src: "a := 1; // foo bar", res: "1"},
 	})
 }
+
+func TestIssue1151(t *testing.T) {
+	type pkgStruct struct{ X int }
+	type pkgArray [1]int
+
+	i := interp.New(interp.Options{})
+	i.Use(interp.Exports{
+		"pkg/pkg": map[string]reflect.Value{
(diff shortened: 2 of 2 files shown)


mvertes:
Thanks for your submission. But I think that a better fix would be to modify `ImportUsed` so that package is added to the universe scope as `<pkg>/_.go` instead of `<pkg>`. The reason for this pattern `<pkg>/_.go` was to avoid possible name space collisions between package names which should be scoped per file, and global var/const/type/fun which are scoped amongst all files in the same package. Still using `<pkg>` in `ImportUsed` is probably a mistake (mine).

firelizzard18:
@mvertes Initially that's what I was going to do. But the normal handling for imports derives the file-specific path from the filename of the current evaluation. For `Eval` this means `<pkg>/_.go`, but I think if I call `EvalPath("foo.go")` that will be `<pkg>/foo.go`. `ImportUsed` is executed outside of any eval call, so we would have to hardcode it to `<pkg>/_.go` or something. Unless I'm wrong about `EvalPath`, that would mean that definitions imported from `ImportUsed` would no longer be accessible to `EvalPath`.

mvertes (approved):
LGTM

PR #1155, opened 2021-06-17
Avoid unnecessary calls to FieldByIndex

### LINKED ISSUE #1149: Yaegi sometimes treats variables of a named array type as structs
I'm working on [gitlab.com/eukano/runway](https://gitlab.com/eukano/runway), in which Yaegi serves as an interpreter for scripts that draw 3D models. The interpreter is set up [here](https://gitlab.com/eukano/runway/-/blob/568c19d470a5694769ec17c9bba7b570b2331b2b/script/interpreter/interpreter.go). An example model/script is [here](https://gitlab.com/eukano/runway/-/blob/568c19d470a5694769ec17c9bba7b570b2331b2b/example/simple.runway).

1. In the interpreter setup, I evaluate `type Vec = geom.Vec`
2. `geom.Vec` is defined as `type Vec = mgl32.Vec`
3. `mgl32.Vec` is defined as `type Vec3 f32.Vec3`
4. `[bug]` In the script, calling a method on a `Vec` causes a panic:

```
reflect: call of reflect.Value.FieldByIndex on array Value
goroutine 1 [running, locked to thread]:
reflect.flag.mustBe(...)
        /usr/lib/go/src/reflect/value.go:221
reflect.Value.FieldByIndex(0x1397ca0, 0xc0006f2e50, 0x191, 0xc0006f2838, 0x0, 0x0, 0xc0006eb600, 0xc0006e8b90, 0x1397ca0)
        /usr/lib/go/src/reflect/value.go:884 +0x305
github.com/traefik/yaegi/interp.getIndexSeqMethod.func2(0xc0002a6210, 0xc0006e8b90)
        REDACTED/src/yaegi/interp/run.go:1996 +0xd8
```

For example, `Vec{1,2,3}.Normalize()` or `var v Vec; v.Normalize()` cause this error. This can be reproduced by cloning Runway and running `go run ./cmd/runway inspect <script>` where the script contains a Vec method call. Compiling/running with `-tags nogtk,noegl` should avoid the CGo depedencies.

Tracing this backwards:

1. The error is raised by `FieldByIndex` because `value(f)` is not a struct

    https://github.com/traefik/yaegi/blob/2a0d29a390dbfa728eb01376d70e62ff9aacb1fd/interp/run.go#L1995-L1998

2. That function is assigned to `n.exec` because the method receiver is not a pointer

    https://github.com/traefik/yaegi/blob/2a0d29a390dbfa728eb01376d70e62ff9aacb1fd/interp/run.go#L1989

3. `getIndexSeqMethod` is assigned to `n.gen` as part of control flow graph generation

   https://github.com/traefik/yaegi/blob/2a0d29a390dbfa728eb01376d70e62ff9aacb1fd/interp/cfg.go#L1589-L1598

firelizzard18:
Fixes #1149

Because of how aliases are handled, `n.gen` is set to `getIndexSeqMethod` or `getIndexSeqPtrMethod` in cases like the one described in #1149. As a result, `FieldByIndex` can be called on a value that is not a struct, which causes a panic. This MR updates those two methods to avoid that call if the index array is empty.

### DIFF
--- a/interp/interp_eval_test.go
+++ b/interp/interp_eval_test.go
@@ -1511,3 +1511,30 @@ func TestIssue1142(t *testing.T) {
 		{src: "a := 1; // foo bar", res: "1"},
 	})
 }
+
+type Issue1149_Array [3]float32
+
+func (v Issue1149_Array) Foo() string  { return "foo" }
+func (v *Issue1149_Array) Bar() string { return "foo" }
+
+func TestIssue1149(t *testing.T) {
+	i := interp.New(interp.Options{})
+	i.Use(interp.Exports{
+		"pkg/pkg": map[string]reflect.Value{
+			"Type": reflect.ValueOf((*Issue1149_Array)(nil)),
+		},
+	})
+	i.ImportUsed()
+
+	_, err := i.Eval(`
+		type Type = pkg.Type
+	`)
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	runTests(t, i, []testCase{
+		{src: "Type{1, 2, 3}.Foo()", res: "foo"},
+		{src: "Type{1, 2, 3}.Bar()", res: "foo"},
+	})
+}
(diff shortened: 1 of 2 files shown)


mvertes (approved):
LGTM

Issue #1139, opened 2021-06-12
Interpreter.Use allows blank import paths


firelizzard18:
This is more of a question, "Should Yaegi behave this way?" But I experienced it as a bug, so I'm entering this as a bug report. #1102 introduced a breaking change to `Interpreter.Use`: previously, an entry like `"pkg": reflect.ValueOf(...)` would add a package named `pkg` with an import path of `pkg`. Now, to get the same effect, the entry should be `"pkg/pkg": ...`.

- **Issue:** Post-#1102, if I `Interpreter.Use(interp.Exports{"pkg": ...})`, the `importPath` of the resulting entry is `"."`.
- **Question**: Does it ever make sense for `importPath` to be `"."`?
- **Solution?**: `Interpreter.Use` panics if `importPath == "."`

I'm using Yaegi to implement a simple scripting language, so I'm restricting what can be imported:

```go
package main

import (
	"reflect"
	"github.com/traefik/yaegi/interp"
	"github.com/traefik/yaegi/stdlib"
)

func main() {
	api := map[string]reflect.Value{ ... }

	I := interp.New()
	I.Use(interp.Exports{
		"fmt":   stdlib.Symbols["fmt"],
		"customPkg": api,
		// ...
	})

	// keys of I.binPkg = "", "."
}
```

mvertes:
`importPath` set to `"."` makes no sense. Panic in that case is one option, another is to silently ignore the entry (instead of adding a key `"."`).  Or let `Use` to return an `error`, at the cost of an API change. We should probably do that while we can (before bump to `v1`). This part is error prone for the user, and a feedback from the function would be useful here.

firelizzard18:
@mvertes I would prefer some form of error. In effect, currently this silently ignores the issue, because it adds a useless entry. If this instead returned an error or panicked, I quickly would have noticed what I did wrong.

Personally, I prefer returning an error, because that allows for other errors in the future, and now is better than post `v1`. If you want to offload the effort, I would be happy to submit a PR. LMK

mvertes:
Yes, go ahead.
Thank you @firelizzard18

mvertes:
Fixed by 1df5dc2

PR #1162, opened 2021-06-22
Update Interpreter.Use API

firelizzard18:
### Background

#1102 changed how `Interpreter.Use` interprets export paths such that the last path component is stripped and used as the package name. This resulted in #1139 - attempting to Use an export with only one path component, such as `foo`, would result in the import path being `.`.

### Breaking API Change

This PR changes the signature of `Interpreter.Use` from `Use(Exports)` to `Use(Exports) error`.

### Fix for #1139

With this PR, if Use is called with an incomplete export path, such as `foo`, Use will return an error.


mvertes (approved):
LGTM

PR #1164, opened 2021-07-01
interp: improve field and method resolution in presence of collisions.

### LINKED ISSUE #1163: Similarly named method signature clash resulting in compile time error
The following program `yaegi_wrong_method.go` triggers a **compile error**:

```go
package main

import "fmt"

type WidgetEvent struct {
        Nothing string
}

type WidgetControl interface {
        HandleEvent(e *WidgetEvent)
}

type Button struct{}

func (b *Button) HandleEvent(e *WidgetEvent) {
}

type WindowEvent struct {
        Something int
}

type Window struct {
        Widget WidgetControl
}

func (w *Window) HandleEvent(e *WindowEvent) {
}

func main() {
        window := Window{
                Widget: &Button{},
        }
        windowevent := &WindowEvent{}
        // The next line uses the signature from the wrong method, resulting in an error.
        // Renaming one of the clashing method names fixes the problem.
        window.HandleEvent(windowevent)
        fmt.Println("OK!")
}
```

Expected result:
```console
$ go run ./yaegi_wrong_method.go
OK!
```

Got:
```console
$ yaegi ./yaegi_wrong_method.go
run: yaegi_wrong_method.go:36:21: cannot use type *main.WindowEvent as type
```

Error started from commit: e32b2ab6bdfc16f225868c979e565eb0340340ff

Workaround: change the name of one of the clashing methods.

mvertes:
The resolution method was not compliant with the Go specification which
requires to retain the object where the field or method is the most
shallowed.

The detection of ambiguous fields or methods (same depth in different
objects) has also been added.

Fixes #1163.

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -1525,6 +1525,18 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 							rtype := n.typ.TypeOf()
 							n.typ = &itype{cat: valueT, rtype: rtype, val: n.typ}
 						}
+					case isStruct(n.typ):
+						// If a method of the same name exists, use it if it is less shallowed.
+						// if method's depth is the same as field's, this is an error.
+						d := n.typ.methodDepth(n.child[1].ident)
+						if d >= 0 && d < len(ti) {
+							goto tryMethods
+						}
+						if d == len(ti) {
+							err = n.cfgErrorf("ambiguous selector: %s", n.child[1].ident)
+							break
+						}
+						fallthrough
 					default:
 						n.gen = getIndexSeq
 						n.typ = n.typ.fieldSeq(ti)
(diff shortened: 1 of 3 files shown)


interp/cfg.go
```diff
@@ -1540,11 +1552,24 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 					// Handle an embedded binary field into a struct field.
 					n.gen = getIndexSeqField
 					lind = append(lind, s.Index...)
+					if isStruct(n.typ) {
+						// If a method of the same name exists, use it if it is less shallowed.
```
mpl, line 1556:
```suggestion
						// If a method of the same name exists, use it if it is shallower than the struct field.
```

interp/cfg.go
```diff
@@ -1511,6 +1511,18 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 			case isStruct(n.typ) || isInterfaceSrc(n.typ):
 				// Find a matching field.
 				if ti := n.typ.lookupField(n.child[1].ident); len(ti) > 0 {
+					if isStruct(n.typ) {
+						// If a method of the same name exists, use it if it is less shallowed.
```
mpl, line 1515:
```suggestion
						// If a method of the same name exists, use it if it is shallower than the struct field.
```

interp/cfg.go
```diff
@@ -1511,6 +1511,18 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
...
+						// if method's depth is the same as field's, this is an error.
+						d := n.typ.methodDepth(n.child[1].ident)
+						if d >= 0 && d < len(ti) {
+							goto tryMethods
```
mpl, line 1519: why not 'goto default', since we want to fallthrough in default anyway? is it forbidden to use goto without a custom label?

mvertes: `default` is not usable as a label in Go, we need a custom label here.  I have not seen a better way to avoid a lot of code duplicates.

mpl: ACK

mpl (approved):

PR #1165, opened 2021-07-01
chore: improve issue templates

ldez:
- use new GitHub issue forms
- update readme badges and links


ldez:
Demo: https://github.com/ldez/gh-issue-templates/issues/new/choose

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,59 @@
+name: Bug Report
+description: Create a report to help us improve
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ⚠️ If you do not respect this template your issue will be closed.
```
mvertes, line 8: Please remove this. I do not think this warning is necessary in this repository so far. I would like to keep as much as possible a minimum level of flexibility and friendliness. In case of (very rare) problem, we can close/suppress the issue anyway.

.github/ISSUE_TEMPLATE/bug_report.yml
```diff
@@ -0,0 +1,59 @@
...
+        ⚠️ Make sure to browse the opened and closed issues before submit your issue.
+
+  - type: textarea
+    id: sample
+    attributes:
+      label: "The following program `sample.go` triggers a panic"
```
mvertes, line 14:
```suggestion
      label: "The following program `sample.go` triggers an unexpected result"
```

.github/ISSUE_TEMPLATE/feature_request.yml
```diff
@@ -0,0 +1,33 @@
+name: Feature request
+description: Propose a change to Yaegi
+
+body:
+  - type: markdown
+    attributes:
+      value: |
+        ⚠️ If you do not respect this template your issue will be closed.
```
mvertes, line 8: Idem as above, please remove this warning.

mvertes (changes requested):

.github/ISSUE_TEMPLATE/bug_report.yml
ldez: This warning was in the previous template and added by you in #333

.github/ISSUE_TEMPLATE/feature_request.yml
ldez: This warning was in the previous template and added by you in #333

.github/ISSUE_TEMPLATE/bug_report.yml
mvertes: Then my 2021 self disagrees with my 2019 self, and right now I stand with my 2021 self (mainly for temporal consistency reasons). Seriously, I think it should be removed, and this PR is an opportunity to do that.

mvertes (approved):
LGTM

PR #1197, opened 2021-07-21
interp: fix map update by assign operators

### LINKED ISSUE #1194: Unary operator ++ does not work on map

```go
package main

func main() {
   m := make(map[int]int)
   m[123]++
   fmt.Println(m)
}
```

### Expected result

```console
map[123:1]
```

### Got

```console
map[]
```

### Yaegi Version

0.9.20

### Additional Notes

_No response_

mvertes:
When operating on map elements, result of assign operators were not
written to the map entry. Now check if the destination of an assign
operator is in a map, and if so, set the result to it.

Fixes #1194.


nrwiersma (approved):
LGTM

PR #1199, opened 2021-07-22
interp: allow for reading source files from diverse filesystems

### LINKED ISSUE #1200: Make use of `fs.FS` (go 1.16 and up) so that source can be loaded from diverse filesystems.
### Proposal

I propose adding a `Filesystem fs.FS` field to `Config` so that the user can supply their own filesystem.

### Background

I use yeagi for my fantasy console, but I also target WASM where there is no default filesystem. Currently I use `Eval(string)` but it means I lose all the benfits of filesystem separation gained from `EvalPath()` resulting in huge monoliths of unseparated code.

As I already have and use an http zipfile `fs.FS` in my fantasy console, being able to set that FS for yaegi allows me to lay out files like a local filesystem (even tho it's dynamically loaded from a zipfile over http) and not have to use `Eval(string)` anymore.

### Workarounds

`Eval(string)`, but I lose all the benefits of file separation that `EvalPath()` gives.

I have started a PR #1199 to discuss how I see implementation (I was curious so I started impl before creating this feature request).

TheMightyGit:
Make use of fs.FS (new to go 1.16) to allow for reading source files from diverse filesystems (local, embed, custom).

* `Options` has a new field `SourcecodeFilesystem fs.FS` so users can supply their own read-only filesystem containing source code.
* Defaults to the local filesystems (via `RealFS` - a thin `os.Open` wrapper complying with `fs.FS`) so regular users should see no change in behaviour.
* When no filesystem is available (e.g. WASM, or if you want to embed files to retain single binary distribution) an alternative filesystem is preferable to using `Eval(string)` as that requires the stringy code to be a single file monolith instead of multiple files. By using an `fs.FS` we can use `EvalPath()` and gain the ability to handle multiple files and packages.
* You can make use of embed filesystems (https://pkg.go.dev/embed) and custom filesystems obeying the `fs.FS` interface (I use one for http served zip files when targeting wasm as there is no local filesystem on wasm). Tests can make use of `fstest.Map`.
* NOTE: This does NOT affect what the running yaegi code considers its local filesystem, this is only for the interpreter finding the source code.

See `example/fs/fs_test.go` for an example.

Fixes #1200.


mvertes:
Thank you for this proposal. I'm happy with the way you are approaching it so far.

> What's the oldest version of go yaegi supports being compiled with? [ CI and docs seem to indicate 1.15/1.16, but the go.mod says 1.12 - which the linter uses to enforce some rules. Master does not compile with 1.14 so I think it's safe to assume 1.15 minimum? Maybe the go.sum should be updated to 1.15? ]

The oldest version of go we support for yaegi is go1.15. We should indeed update go.sum to this version. Note that we will bump it to go1.16 as soon as go1.17 is released.

Regarding the polyfill of go1.15, as release of go1.17 should come quite soon now (in august, according to https://github.com/golang/go/milestones), we should probably wait for it and skip the polyfill entirely). We usually update yaegi quickly after a new Go version is release anyway. In the mean time we could keep this PR opened if necessary.

> Does this change affect what running yaegi code considers its own filesystem? I assume not... but needs checking.

As far as I see, this change does not affect what yaegi considers as its filesystem, or at least it should not. But the virtualisation of FS through embed is something that we would certainly consider in the sandboxing layer soon after this PR. Note: at this moment only os.Stdin, os.Stdout and os.Stderr file descriptors are sandboxed.

> Maybe I should rename the option ReadonlySourcecodeFilesystem fs.FS to hammer the point home?

this option name, although a bit too long for my taste, should be fine.

> When I run make tests locally on my mac tests fails 28 tests on master - is this expected?

Will look at it on my side.

TheMightyGit:
@mvertes I think waiting for 1.17 is a good idea, the ugliest part of this PR is the 1.15 polyfill and it seems silly to add it now only to remove it soon. So I'll remove the polyfill code from this PR and it can sit open until 1.17 support lands and 1.15 is retired. Thank you for your time and efforts <3

mvertes:
@TheMightyGit I think this PR is almost ready for merge. Is it ok for you to remove the draft status, and maybe provide additional documentation about the new filesystem config option, if you feel so. Thanks.

mvertes (approved):
LGTM

PR #1210, opened 2021-08-02
interp: fix package lookup for type analysis of local types

### LINKED ISSUE #1203: d16bd4bc breaks kortschak/yaegiconf
### The following triggers an unexpected result

This bug is not a use of the interpreter as an executable, but as an embedded interpreter. So this field is not informative.

Attempt to run tests for github.com/kortschak/yaegiconf at or after d16bd4bcdb8685f7ce88a0ee2261f7d945a6cc84 fail. This can be seen here by running the tests on this branch https://github.com/kortschak/yaegiconf/tree/broken.

```go
package yaegiconf_test

import (
	"context"
	"fmt"
	"log"
	"reflect"

	"github.com/kortschak/yaegiconf"
)

func Example_struct() {
	type Config struct {
		N int
		F float64
	}
	var c Config
	err := yaegiconf.EvalTo(&c, `config.Value{N: 5, F: 0.1}`)
	if err != nil {
		log.Fatalf("failed to parse configuration: %v", err)
	}

	fmt.Printf("%#v\n", c)

	// Output:
	//
	// yaegiconf_test.Config{N:5, F:0.1}
}
```

### Expected result

```console
$ go test            
PASS
ok  	github.com/kortschak/yaegiconf	10.004s
```

### Got

```console
$ go test                               
2021/07/26 10:11:44 failed to parse configuration: 1:21: import "config" error: unable to find source related to: "config". Either the GOPATH environment variable, or the Interpreter.Options.GoPath needs to be set
exit status 1
FAIL	github.com/kortschak/yaegiconf	0.002s
```

### Yaegi Version

d16bd4bcdb8685f7ce88a0ee2261f7d945a6cc84

### Additional Notes

It's pretty clear that I'm abusing package namespaces here, but it did previously work. I don't see any information in #1102 that shows how I should alter the use to make something like this work (and I suspect that the valid changes won't because this is running in a \_test package.

This is probably a documentation issue more than a correctness issue.

kortschak:
Fixes #1203.

### DIFF
--- a/interp/gta.go
+++ b/interp/gta.go
@@ -1,6 +1,7 @@
 package interp
 
 import (
+	"path"
 	"path/filepath"
 	"reflect"
 )
@@ -187,6 +188,9 @@ func (interp *Interpreter) gta(root *node, rpath, importPath string) ([]*node, e
 			}
 			// Try to import a binary package first, or a source package
 			var pkgName string
+			if packageName := path.Base(ipath); path.Dir(ipath) == packageName {
+				ipath = packageName
+			}
 			if pkg := interp.binPkg[ipath]; pkg != nil {
 				switch name {
 				case "_": // no import of symbols


mvertes (approved):
LGTM

PR #1216, opened 2021-08-14
feat: fix zero instance issue

### LINKED ISSUE #1215: Instantiating empty object in function param is nil

```go
package main

import (
	"fmt"
)

type data struct {
	S string
}

func render(v interface{}) {
	fmt.Println(v)
}

func main() {
	render(data{})
}
```

### Expected result

```console
$ go run ./sample.go
{}
```

### Got

```console
$ go run ./sample.go
<nil>
```

### Yaegi Version

v0.9.21, 37fe3422d8de70e793e1c52b8364721afca3b764

### Additional Notes

_No response_

nrwiersma:
When calling a function, if the input param is a "zero" instance, it is not set on the input. This is an issue where the param is an `interface{}` as a `nil` value is set instead of the zero value.

The actual solution for this is to remove the `if !val.IsZero()`, this however runs into an issue on `_test/struct48.go` where we have a zero recursive struct instance (`*interface{}`) and we have no way to get its real type. Should a way be figured out to keep tabs on the original type, the `if` can go away, and in the zero case of `genValueRecursiveInterfacePtrValue`, the actual zero type can be returned. 

Fixes #1215

### DIFF
--- a/_test/struct60.go
+++ b/_test/struct60.go
@@ -1,0 +1,20 @@
+package main
+
+import (
+	"fmt"
+)
+
+type data struct {
+	S string
+}
+
+func render(v interface{}) {
+	fmt.Println(v)
+}
+
+func main() {
+	render(data{})
+}
+
+// Output:
+// {}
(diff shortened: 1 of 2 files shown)


mvertes (approved):
LGTM

PR #1217, opened 2021-08-14
interp: fix handling of forward function declaration

### LINKED ISSUE #1214: Unexpected panic when calling a function

```go
package main

import (
	"fmt"
	"sync"
)

type eventState struct {
	EntityID string `json:"entity_id"`
	State    string `json:"state"`
}

type eventChanged struct {
	EventType string `json:"event_type"`
	Data      struct {
		NewState eventState `json:"new_state"`
	} `json:"data"`
}

type State struct {
	ID    string
	State *bool
}

func main() {
	var wg sync.WaitGroup
	wg.Add(1)

	ch := make(chan State, 10)
	go func() {
		defer wg.Done()

		for s := range ch {
			fmt.Println(s.ID)
		}
	}()

	wg.Add(1)
	go func() {
		defer wg.Done()

		event := eventChanged{
			EventType: "state_changed",
			Data: struct {
				NewState eventState `json:"new_state"`
			}{
				NewState: eventState{
					EntityID: "test",
					State:    "on",
				},
			},
		}

		ch <- State{
			ID:    event.Data.NewState.EntityID,
			State: stateToBool(event.Data.NewState.State),
		}

		close(ch)
	}()

	wg.Wait()
}

func stateToBool(state string) *bool {
	switch {
	case state == "on" || state == "open":
		return boolPtr(true)
	case state == "off" || state == "closed":
		return boolPtr(false)
	}

	return nil
}

func boolPtr(b bool) *bool {
	return &b
}
```

### Expected result

```console
$ go run ./sample.go
test
```

### Got

```console
$ go run ./sample.go
./sample.go:56:11: panic
panic: reflect: call of reflect.Value.Call on ptr Value [recovered]
	panic: reflect: call of reflect.Value.Call on ptr Value

goroutine 21 [running]:
github.com/traefik/yaegi/interp.runCfg.func1(0xc0001cee70, 0xc000499b00, 0xc00006cf98)
	/Users/nrwiersma/Go/src/github.com/traefik/yaegi/interp/run.go:184 +0x253
panic(0x1a513e0, 0xc000191320)
	/usr/local/go/src/runtime/panic.go:965 +0x1b9
reflect.flag.mustBe(...)
	/usr/local/go/src/reflect/value.go:221
reflect.Value.Call(0x1a01100, 0xc000223c60, 0x16, 0xc000191308, 0x1, 0x1, 0x21ea940, 0x0, 0x0)
	/usr/local/go/src/reflect/value.go:335 +0x174
github.com/traefik/yaegi/interp.call.func6(0xc0001cee70, 0xc0004ac070)
	/Users/nrwiersma/Go/src/github.com/traefik/yaegi/interp/run.go:1177 +0x10c7
github.com/traefik/yaegi/interp.runCfg(0xc000499b00, 0xc0001cee70)
	/Users/nrwiersma/Go/src/github.com/traefik/yaegi/interp/run.go:190 +0x87
created by github.com/traefik/yaegi/interp.call.func6
	/Users/nrwiersma/Go/src/github.com/traefik/yaegi/interp/run.go:1277 +0xd68
exit status 2
```

### Yaegi Version

v0.9.21, 37fe3422d8de70e793e1c52b8364721afca3b764

### Additional Notes

Moving the functions `stateToBool` and `boolPtr` above `main` stops the panic.

nrwiersma:
Set node address in `val` field at creation of `funcDecl` node so it can be used correctly at closure generation, even in the case of forward function declarations, where the value was zero.

Fixes #1214

### DIFF
--- a/interp/value.go
+++ b/interp/value.go
@@ -197,7 +197,19 @@ func genValue(n *node) func(*frame) reflect.Value {
 		} else {
 			v = reflect.ValueOf(n.val)
 		}
-		return func(f *frame) reflect.Value { return v }
+		return func(f *frame) reflect.Value {
+			// It is possible that the nodes val had not been set
+			// when getting the value. In this case, recheck it
+			// at runtime.
+			if !v.IsValid() {
+				return v
+			}
+
+			if w, ok := n.val.(reflect.Value); ok {
+				return w
+			}
+			return reflect.ValueOf(n.val)
+		}
 	default:
 		if n.rval.IsValid() {
 			convertConstantValue(n)
--- a/interp/value.go
+++ b/interp/value.go
@@ -197,19 +197,17 @@ func genValue(n *node) func(*frame) reflect.Value {
 		} else {
 			v = reflect.ValueOf(n.val)
 		}
-		return func(f *frame) reflect.Value {
-			// It is possible that the nodes val had not been set
-			// when getting the value. In this case, recheck it
-			// at runtime.
-			if !v.IsValid() {
-				return v
-			}
-
-			if w, ok := n.val.(reflect.Value); ok {
-				return w
-			}
-			return reflect.ValueOf(n.val)
-		}
+	  // If node value is not yet resolved due to forward function declaration,
+        // then we have to perform node resolution at runtime.
+        if _, ok := v.Interface().(*node); ok {
+            return func(f *frame) reflect.Value { return v }
+        }
+        return func(f *frame) reflect.Value {
+            if w, ok := n.val.(reflect.Value); ok {
+                return w
+            }
+            return reflect.ValueOf(n.val)
+        }
 	default:
 		if n.rval.IsValid() {
 			convertConstantValue(n)
--- a/interp/value.go
+++ b/interp/value.go
@@ -197,17 +197,7 @@ func genValue(n *node) func(*frame) reflect.Value {
 		} else {
 			v = reflect.ValueOf(n.val)
 		}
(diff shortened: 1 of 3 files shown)


interp/value.go
```diff
@@ -197,7 +197,19 @@ func genValue(n *node) func(*frame) reflect.Value {
...
+			// at runtime.
+			if !v.IsValid() {
+				return v
+			}
+
+			if w, ok := n.val.(reflect.Value); ok {
+				return w
+			}
+			return reflect.ValueOf(n.val)
+		}
```
mvertes, line 212:
```suggestion
	  // If node value is not yet resolved due to forward function declaration,
        // then we have to perform node resolution at runtime.
        if _, ok := v.Interface().(*node); ok {
            return func(f *frame) reflect.Value { return v }
        }
        return func(f *frame) reflect.Value {
            if w, ok := n.val.(reflect.Value); ok {
                return w
            }
            return reflect.ValueOf(n.val)
        }
```

mvertes (changes requested):
Your fix can be slightly improved to preserve speed in the best case (no forward function declarations), otherwise it looks ok to me.

mvertes:
After closer examination, it is not necessary to hack `genValue`. a better fix is to set correctly `n.val` for `funcDecl` node once at creation. forward function declarations should be not a concern when generating closures, as all necessary symbols must be correctly populated at this stage. I will commit my fix in this PR in a minute.

mvertes (approved):
LGTM

PR #1220, opened 2021-08-17
stdlib: support go1.17 unsafe functions

nrwiersma:
This adds support for go1.17 `unsafe.Add` and `unsafe.Slice`.


mvertes (approved):
LGTM

PR #1226, opened 2021-08-24
interp: fix append with 1 argument

### LINKED ISSUE #1224: Builtin append with only one argument fails

```go
package main

func main() {
	var C []int
	_ = append(C)
}
```

### Expected result

```console
$ go run ./sample.go
<no error>
```

### Got

```console
$ yaegi ./sample.go
run: ./main.go:5:6: CFG post-order panic: runtime error: index out of range [1] with length 1
```

### Yaegi Version

v0.9.23

### Additional Notes

In addition to the above error `append(nil)` fails with a nil-dereference error where the go compiler gives a suitable compiler error.

mvertes:
Fixes #1224.

### DIFF
--- a/interp/interp_eval_test.go
+++ b/interp/interp_eval_test.go
@@ -496,6 +496,7 @@ func TestEvalSliceExpression(t *testing.T) {
 		{src: `a := []int{0,1,2,3}[1:3:]`, err: "1:51: 3rd index required in 3-index slice"},
 		{src: `a := []int{0,1,2}[3:1]`, err: "invalid index values, must be low <= high <= max"},
 		{src: `a := []int{0,1,2}; append(a)`, res: "[0 1 2]"},
+		{src: `append(nil)`, err: "first argument to append must be slice; have nil"},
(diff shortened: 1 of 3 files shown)


interp/interp_eval_test.go
```diff
@@ -495,6 +495,7 @@ func TestEvalSliceExpression(t *testing.T) {
 		{src: `a := []int{0,1,2,3}[1::4]`, err: "1:49: 2nd index required in 3-index slice"},
 		{src: `a := []int{0,1,2,3}[1:3:]`, err: "1:51: 3rd index required in 3-index slice"},
 		{src: `a := []int{0,1,2}[3:1]`, err: "invalid index values, must be low <= high <= max"},
+		{src: `a := []int{0,1,2}; append(a)`, res: "[0 1 2]"},
```
nrwiersma, line 498: perhaps this can move to `TestEvalBuiltin` with the other `append` tests. We should also add a test for `append(nil)`, WDYT?

nrwiersma (changes requested):

interp/interp_eval_test.go
mvertes: yes, done

nrwiersma (approved):
LGTM

Issue #1228, opened 2021-08-25
TestYaegiCmdCancel is failing on non 48-bit VMA


creekorful:
### The following program `sample.go` triggers an unexpected result

```go
yaegi := filepath.Join(tmp, "yaegi")
build := exec.Command("go", "build", "-race", "-o", yaegi, ".") # -race is only supported for 48-bit VMA
out, err := build.CombinedOutput()
```

### Expected result

```console
test should pass
```

### Got

```console
test has failed
```

### Yaegi Version

0.9.21

### Additional Notes

Bug has been reported in Debian: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=992941

creekorful:
Here's the patch. I've tried building yaegi again on armhf and it looks like the build is passing :)
I'll upload a new version of yaegi on the Debian archive to see if it solves the issue on all arches.

Cheers,

PR #1229, opened 2021-08-25
test: only enable race detector when supported

creekorful:
Closes: #1228


creekorful:
> @creekorful could you please rebase this PR ?
> Now that #1232 is merged, I expect the lint error to be fixed, and build and tests on all architectures to be fixed, although I could only test a few. If you could confirm that on your side, it would be very useful.
> Thanks!

@mvertes Looks like your fixes solve the issue ([see the builds](https://buildd.debian.org/status/package.php?p=golang-github-traefik-yaegi)).
Many thanks!

PR #1236, opened 2021-08-30
interp: create real recursive types with unsafe type swapping

nrwiersma:
As the unsafe and pointer methods in `reflect` are to be depreciated, and seeing no replacement functions, it is now forced that some unsafe is needed to replace this as when and interface is dereferenced it is made unsettable by reflect.

With this in mind, this adds real recursive types by hot swapping the struct field type on the fly. This removes a lot of compensation code, simplifying all previous cases.

**Note:** While the struct field type is swapped for the real type, the type string is not changed. Due to this, unsafe will recreate the same type.

### DIFF
--- a/internal/unsafe2/unsafe.go
+++ b/internal/unsafe2/unsafe.go
@@ -1,0 +1,52 @@
+package unsafe2
+
+import (
+	"reflect"
+	"unsafe"
+)
+
+type dummy struct{}
+
+// DummyType represents a stand-in for a recursive type.
+var DummyType = reflect.TypeOf(dummy{})
+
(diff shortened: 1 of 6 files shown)


mvertes:
The following test, ok with `v0.9.23`, fails here with a panic, but also probably since #1231.

```go
package main

func main() {
    b := []byte{1, 3, 50}
    w := uint32(b[1])
    println(w)
}   

// Output:
// 3
```

mvertes:
The test is above is in fact ok. Sorry for the noise

mvertes (approved):
LGTM

PR #1239, opened 2021-08-31
interp: fix a regression in type checks

### LINKED ISSUE #1238: yaegi test golang.org/x/crypto/bcrypt fails at parse

```go
$ yaegi test -v golang.org/x/crypto/bcrypt
```

### Expected result

```console
$ yaegi test -v golang.org/x/crypto/bcrypt
== RUN   TestCost
-- PASS: TestCost (0.00s)
[...]
```

### Got

```console
$ yaegi test -v golang.org/x/crypto/bcrypt
test: /home/marc/go/src/golang.org/x/crypto/bcrypt/bcrypt.go:135:12: cannot use type untyped rune as type uint8 in assignment
```

### Yaegi Version

master (after #1231)

### Additional Notes

This was working (minus errors for some tests) in v0.9.23.

mvertes:
When type checking assignments or conversions, we may have to check
if a constant value, known at compile time, is representable in a given type.
It involves retrieving the value from the node, which is not always
populated in type object. Before #1231, the node value was populated
correctly in the case of #1238, but not since. Anyway, there is no guarantee that this value is
always set, depending on how the type is built.

The assignableTo type method now requires the node
value to be passed explicitely, not relying on how the type was
obtained.

Fixes #1238.

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -540,6 +540,7 @@ func (interp *Interpreter) cfg(root *node, importPath string) ([]*node, error) {
 					sym, level, _ = sc.lookup(dest.ident)
 				}
 
+				log.Println(n.cfgErrorf("assign"))
 				err = check.assignExpr(n, dest, src)
 				if err != nil {
 					break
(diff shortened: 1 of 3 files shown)


mvertes:
closed in favour of #1240

Issue #1158, opened 2021-06-19
panic when import `golang.org/x/text`


WLun001:
The following program `sample.go` triggers a panic:

```go
package main

import (
	"golang.org/x/text/language"
	"log"
)

func main() {
	log.Println(language.English)
}

```

Expected result:
```console
$ go run ./sample.go

2021/06/19 21:42:46 en
```

Got: `panic run: language: tag is not well-formed`
 Links to panic line number
- [golang.org/x/text/internal/language/tags.go:12](https://github.com/golang/text/blob/master/internal/language/tags.go#L12)
- [golang.org/x/text/internal/language/compact/compact.go:59](https://github.com/golang/text/blob/master/internal/language/compact/compact.go#L59)

```console
$ yaegi ./sample.go


/Users/wlun/go/src/golang.org/x/text/internal/language/tags.go:12:3: panic
/Users/wlun/go/src/golang.org/x/text/internal/language/compact/compact.go:59:20: panic
run: language: tag is not well-formed
goroutine 1 [running]:
runtime/debug.Stack(0x1, 0xc0013c6800, 0x40)
	/usr/local/Cellar/go/1.16.3/libexec/src/runtime/debug/stack.go:24 +0x9f
github.com/traefik/yaegi/interp.(*Interpreter).eval.func1(0xc00223fc18)
	/Users/wlun/go/pkg/mod/github.com/traefik/yaegi@v0.9.18/interp/interp.go:503 +0xc8
panic(0x1b3a800, 0xc001a3dc20)
	/usr/local/Cellar/go/1.16.3/libexec/src/runtime/panic.go:965 +0x1b9
github.com/traefik/yaegi/interp.runCfg.func1(0xc001bccfd0, 0xc002052100, 0xc00223de68)
	/Users/wlun/go/pkg/mod/github.com/traefik/yaegi@v0.9.18/interp/run.go:185 +0x253
panic(0x1b3a800, 0xc001a3dc20)
	/usr/local/Cellar/go/1.16.3/libexec/src/runtime/panic.go:965 +0x1b9
github.com/traefik/yaegi/interp.runCfg.func1(0xc001bcd130, 0xc001295900, 0xc00223dc58)
	/Users/wlun/go/pkg/mod/github.com/traefik/yaegi@v0.9.18/interp/run.go:185 +0x253
panic(0x1b3a800, 0xc001a3dc20)
[...]

PR #1243, opened 2021-09-03
interp: fix parsing of late binding consts

nrwiersma:
When a const is late binding and specified with a type, the GTA defineStmt was creating the symbol with the current scopes `iota` which is incorrect. The symbol should be created with the source nodes `rval`.

Related to #1158


mvertes (approved):
LGTM

PR #1244, opened 2021-09-04
interp: types should not recover data for aliases

nrwiersma:
When `nodeType` recovers names and methods, it can overwrite the data if the type is an aliasT. When aliasing a type, do not recover the methods, this will be done in the GTA typeSpec pass.

Related to #1158


mvertes (approved):
LGTM

PR #1254, opened 2021-09-16
interp: allow GTA revisit for unresolved define statements

### LINKED ISSUE #1253: GTA does not allow revisit of define types

```go
package main

import (
	"fmt"

	"github.com/traefik/yaegi/_test/method38"
)

func main() {
	fmt.Println(method38.Get())
}

```

`method38/a.go`
```go
package method38

import "sync"

func NewPool() Pool { return Pool{} }

type Buffer struct {
	bs   []byte
	pool Pool
}

type Pool struct {
	p *sync.Pool
}

var (
	_pool = NewPool()
	Get   = _pool.Get
)

```

`method38/b.go`

### Expected result

```console
$  go run .sample.go
&{[] {<nil>}}
```
`method38/b.go`
```go
package method38

func (p Pool) Get() *Buffer { return &Buffer{} }

```

### Got

```console
$ yaegi run ./sample.go
run: ./sample.go:6:2: import "github.com/traefik/yaegi/_test/method38" error: github.com/traefik/yaegi/_test/method38/a.go:12:11: undefined selector Get
```

### Yaegi Version

b591ba0e78db7557b5d4e06cc42970d702ce0728

### Additional Notes

_No response_

nrwiersma:
When a define statement relies on a selector or type that may exist in another file it should revisit once GTA is complete. This allows that revisit.

**Note:** In order to keep the original GTA error for the define statement, so the error received is correct and meaningful, I have added a node `meta` property. I decided to make it generic as it may be useful in future. There may be a better way to stash errors across the GTA runs and am open to suggestion here.

Fixes #1253

### DIFF
--- a/_test/method38.go
+++ b/_test/method38.go
@@ -1,0 +1,14 @@
+package main
+
+import (
+	"fmt"
+
+	"github.com/traefik/yaegi/_test/method38"
+)
+
+func main() {
+	fmt.Println(method38.Get())
+}
+
+// Output:
+// &{[] {<nil>}}
--- a/_test/method38/a.go
+++ b/_test/method38/a.go
@@ -1,0 +1,19 @@
+package method38
+
+import "sync"
+
+func NewPool() Pool { return Pool{} }
+
+type Buffer struct {
+	bs   []byte
+	pool Pool
+}
+
+type Pool struct {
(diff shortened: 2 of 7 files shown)


mvertes (approved):
LGTM

PR #1257, opened 2021-09-20
interp: support type spec assign

### LINKED ISSUE #1154: interpreter allows declaring methods on non-local types when Go does not
With the following program:

```go
package main

import "fmt"

type Foo = int

func (f Foo) Bar() int { return int(f * f) }

func main() {
	x := Foo(1)
	fmt.Println(x.Bar())
}
```

Go allows methods on types declared as `type Foo int`, but with the `=` it considers `Foo` to be 'non-local' and does not allow methods to be defined.

Expected result:
```console
$ go run ./test.go 
# command-line-arguments
./test.go:7:6: cannot define new methods on non-local type int
./test.go:11:15: x.Bar undefined (type int has no field or method Bar)
```

Got (on master):
```console
$ go run ./cmd/yaegi/ test.go 
1
```

nrwiersma:
This supports type spec assign types which disallows the attachment of methods onto the alias type as per the Go spec.

Fixes #1154

### DIFF
--- a/_test/type27.go
+++ b/_test/type27.go
@@ -1,0 +1,17 @@
+package main
+
+import "fmt"
+
+type Foo = int
+
+func (f Foo) Bar() int {
+	return f * f
+}
+
+func main() {
+	x := Foo(1)
+	fmt.Println(x.Bar())
+}
+
+// Error:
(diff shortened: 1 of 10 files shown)


mvertes (approved):
LGTM

PR #1261, opened 2021-09-21
interp: add a function to directly compile Go AST

### LINKED ISSUE #1251: interp: add a function to compile AST (instead of source)
### Proposal

```go
func (interp *Interpreter) CompileAST(ast.Node) (*Program, error)
```

This would accept Go AST provided by the user. Currently, the two supported pathways, compile a file and compile a function body, are equivalent to passing in an `*ast.File` or an `*ast.BlockStmt`, respectively. The new function could explicitly support only a File or BlockStmt input, though the existing `(*Interpreter).ast` may handle arbitrary (valid) ASTs.

I wrote a relatively small [patch](https://github.com/traefik/yaegi/compare/master...firelizzard18:compile-ast) for this. Would you accept this as a PR?

### Background

I'm playing around with writing a Go-/Yaegi-based shell. I want to support string interpolation and other features foreign to Go. This would require manipulation of the AST. While I could generate source, I would much prefer to directly pass the modified AST to Yaegi.

### Workarounds

Parsing the source, modifying the AST, writing it back out to a buffer, and handing that buffer to Yaegi, which would parse the source again.

firelizzard18:
Adds CompileAST, which can be used to compile Go AST directly. This
allows users to delegate parsing of source to their own code instead of
relying on the interpreter.

CLoses #1251

### DIFF
--- a/interp/ast.go
+++ b/interp/ast.go
@@ -1,7 +1,6 @@
 package interp
 
 import (
-	"errors"
 	"fmt"
 	"go/ast"
 	"go/constant"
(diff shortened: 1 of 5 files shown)


mvertes (approved):
LGTM

PR #1263, opened 2021-09-24
interp: handle alias of an alias

nrwiersma:
When dealing with an alias of an alias, the actual underlying type of the source alias should be used.


mvertes (approved):
LGTM

Issue #1259, opened 2021-09-21
GTA/CFG on TypeSpec should always result in an `aliasT`


nrwiersma:
### Proposal

Go spec does not allow for non aliasT types to have methods, except for a struct through composition. This is not entirely implemented in yaegi. 
1. When GTA runs a typespec it does not always result in an `aliasT` (which should probably actually be called a `namedT`). 
2. An `aliasT` should then never copy the methods of the underlying type except for a struct type through composition.

The second point would be simpler to implement if a `structT` could only have methods through composition, and `namedOf` could then offload all intermediate `aliasT`s.

### Background

Make it simpler to reason about type and method inheritance to more faithfully implement the Go spec.

### Workarounds

None that I know of.

PR #1264, opened 2021-09-24
interp: handle recursive and incomplete seen types in nodeType

nrwiersma:
This comes from experiments looking into #1259 where incomplete twice seen types are marked as complete. To mitigate the problem instead of a map of seen types in `nodeType` a slice is used as a cheap way to keep track of our current path through the node tree.


mvertes (approved):
LGTM

PR #1267, opened 2021-09-29
Dev

Bai-Yingjie:
This PR includes 2 small features I am working on:
1. Add a tag option in the extract program to optionally give an opportunity to tailor some packages in for example the stdlib, for people who is interested in keeping the final binary size as smaller as possible.

Below `go generate`
```
//go:generate ../cmd/extract/extract -name stdlib -tag stdmime mime mime/multipart mime/quotedprintable
````
produces a header to `stdlib/mime-multipart.go`
```
// Code generated by 'yaegi extract mime/multipart'. DO NOT EDIT.

// +build go1.16,!go1.17,stdmime
```

2. In some cases, we use many yaegi instances in one process, each interpreter will expect its own os.Args. This is done in `interp.go` as an option in Options besides Stdin Stdout and Stderr.

### DIFF
--- a/cmd/yaegi/extract.go
+++ b/cmd/yaegi/extract.go
@@ -18,12 +18,14 @@ func extractCmd(arg []string) error {
 	var name string
 	var exclude string
 	var include string
+	var tag string
 
 	eflag := flag.NewFlagSet("run", flag.ContinueOnError)
 	eflag.StringVar(&licensePath, "license", "", "path to a LICENSE file")
 	eflag.StringVar(&name, "name", "", "the namespace for the extracted symbols")
 	eflag.StringVar(&exclude, "exclude", "", "comma separated list of regexp matching symbols to exclude")
 	eflag.StringVar(&include, "include", "", "comma separated list of regexp matching symbols to include")
+	eflag.StringVar(&tag, "tag", "", "build tag to be added to the created package")
 	eflag.Usage = func() {
 		fmt.Println("Usage: yaegi extract [options] packages...")
 		fmt.Println("Options:")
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: cmd/yaegi/extract.go [hunk 1]
```diff
@@ -18,12 +18,14 @@ func extractCmd(arg []string) error {
...
 	var exclude string
 	var include string
+	var tag string
 
 	eflag := flag.NewFlagSet("run", flag.ContinueOnError)
 	eflag.StringVar(&licensePath, "license", "", "path to a LICENSE file")
 	eflag.StringVar(&name, "name", "", "the namespace for the extracted symbols")
 	eflag.StringVar(&exclude, "exclude", "", "comma separated list of regexp matching symbols to exclude")
 	eflag.StringVar(&include, "include", "", "comma separated list of regexp matching symbols to include")
+	eflag.StringVar(&tag, "tag", "", "build tag to be added to the created package")
```
mvertes, line 28: It would be useful to specify multiple tags at once. It could be achieved by managing `tag` as a comma separated list, as for `include` or `exclude` options.

interp/interp.go
```diff
@@ -181,6 +181,7 @@ type opt struct {
 	stdin        io.Reader     // standard input
 	stdout       io.Writer     // standard output
 	stderr       io.Writer     // standard error
+	args         []string      // cmdline args
```
mvertes, line 184: The addition of `args` to `opt` is a completely different feature. It should not be mixed in the same PR and should submitted in a separate one. Thanks.

mvertes (changes requested):
Thanks for your submission. I'm ok with the 2 features you propose but please do not mix them and submit one separate PR per feature.

mvertes:
closed in favour of #1277 and #1278

PR #1277, opened 2021-10-07
interp: add args option to let each interpreter have seperate args

Bai-Yingjie:
In some cases, we use many yaegi instances in one process, each interpreter will expect its own os.Args. This is done in interp.go as an option in interp.Options besides Stdin Stdout and Stderr.


Bai-Yingjie:
Test added, thanks for pointing it out.

PR #1295, opened 2021-10-20
interp: virtualize environment in restricted mode

mvertes:
In restricted mode, replace environment related symbols in
stdlib os package by a version which operates on a private copy
per interpreter context.

It allows to have concurrent interpreters in the same process
operating each in their own environment without affecting each
other or the host.

If unrestricted opt is set, this behaviour is disabled, and the
default symbols from stdlib are used.

Note also that no modification is done for syscall package, as it
should be not used in restricted mode.

### DIFF
--- a/cmd/yaegi/run.go
+++ b/cmd/yaegi/run.go
@@ -47,7 +47,12 @@ func run(arg []string) error {
 	}
 	args := rflag.Args()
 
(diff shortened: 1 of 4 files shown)


ldez (approved):
LGTM

PR #1305, opened 2021-10-28
interp: fix type processing to support multiple recursive fields 

### LINKED ISSUE #1304: multiple recursive fields in a struct

```go
package main

type Node struct {
    Name  string
    Alias *Node
    Child []*Node
}   

func main() {
    n := &Node{Name: "parent"}
    n.Child = append(n.Child, &Node{Name: "child"})
    println(n.Name, n.Child[0].Name)
}
```

### Expected result

```console
$ go run ./sample.go
parent child
```

### Got

```console
$ yagi ./sample.go
./sample.go:11:28: cannot use type *main.Node as type *unsafe2.dummy
```

### Yaegi Version

v0.10.0

### Additional Notes

This issue is related to #1296.

Note also that if we comment the unused field `Alias *Node` in type definition, the test works as expected.

mvertes:
Fixes #1304


interp/type.go
```diff
@@ -1632,6 +1655,7 @@ type fieldRebuild struct {
 type refTypeContext struct {
 	defined    map[string]*itype
```
mpl, line 1656:
```suggestion
	// defined keeps track of which itypes we've already seen, so we can detect
	// recursive types. keyed by type name.
	defined    map[string]*itype
```

interp/type.go
```diff
@@ -1632,6 +1655,7 @@ type fieldRebuild struct {
...
 	refs       map[string][]fieldRebuild
```
mpl, line 1657:
```suggestion

	// refs keeps track of all the places (in the same type recursion) where the
	// type name (as key) is used as a field of another (or possibly the same) struct
	// type. Each of these fields will then live as an unsafe2.dummy type until the
	// whole recursion is fully resolved, and the type is fixed.
	refs       map[string][]fieldRebuild
```

interp/type.go
```diff
@@ -1632,6 +1655,7 @@ type fieldRebuild struct {
...
+	rect       *itype
```
mpl, line 1658:
```suggestion
	// when we detect for the first time that we are in a recursive type (thanks to
	// defined), we keep track of the first occurence of the type where the recursion
	// started, so we can restart the last step that fixes all the types from the same
	// "top-level" point.
	rect       *itype
```

interp/type.go
```diff
@@ -1672,9 +1741,10 @@ func (t *itype) refType(ctx *refTypeContext) reflect.Type {
 			return dt.rtype
 		}
 
-		// To indicate that a rebuild is needed on the nearest struct
-		// field, create an entry with a nil type.
+		// To indicate that a rebuild is needed on the englobing struct,
```
mpl, line 1744:
```suggestion
		// The recursion has not been fully resolved yet.
		// To indicate that a rebuild is needed on the englobing struct,
```

interp/type.go
```diff
@@ -1672,9 +1741,10 @@ func (t *itype) refType(ctx *refTypeContext) reflect.Type {
...
+		// return a dummy field type and create an entry with an empty fieldRebuild.
 		flds := ctx.refs[name]
+		ctx.rect = dt
 		ctx.refs[name] = append(flds, fieldRebuild{})
```
mpl, line 1748:
```suggestion
		// we know we are used as a field by someone, but we don't know by who
		// at this point in the code, so we just mark it as an empty fieldRebuild for now.
		// We'll complete the fieldRebuild in the caller.
		ctx.refs[name] = append(flds, fieldRebuild{})
```

interp/type.go
```diff
@@ -1733,12 +1802,22 @@ func (t *itype) refType(ctx *refTypeContext) reflect.Type {
...
+					if strings.HasSuffix(f.Type.String(), "unsafe2.dummy") {
+						unsafe2.SetFieldType(s.rtype, i, ctx.rect.fixDummy(s.rtype.Field(i).Type))
+					}
+				}
+			}
+		}
 
 		// The rtype has now been built, we can go back and rebuild
 		// all the recursive types that relied on this type.
 		for _, f := range ctx.refs[name] {
```
mpl, line 1818:
```suggestion
		// However, as we are keyed by type name, if two or more (recursive) fields at
		// the same depth level are of the same type, they "mask" each other, and only one
		// of them is in ctx.refs, which means this pass below does not fully do the job.
		// Which is why we have the pass above that is done one last time, for all fields,
		// one the recursion has been fully resolved.
		for _, f := range ctx.refs[name] {
```

interp/type.go
```diff
@@ -1733,12 +1802,22 @@ func (t *itype) refType(ctx *refTypeContext) reflect.Type {
 			}
 		}
 		t.rtype = reflect.StructOf(fields)
+		if ctx.isComplete() {
+			for _, s := range ctx.defined {
```
mpl, line 1806:
```suggestion
			// Final pass after the recursion has been fully resolved, to fix one last time
			// all the struct fields that were dummy types and that had been "masked" by their
			// siblings. See explanation on block below.
			for _, s := range ctx.defined {
```

mpl (approved):

PR #1329, opened 2021-12-13
interp: fix derivation of type of slice expression of binary object

### LINKED ISSUE #1328: slice expression

```go
package main

import (    
    "crypto/sha1"
    "encoding/hex"                                                                                    
)                    
                                      
func main() {                           
    script := "hello"                                                                                 
    sumRaw := sha1.Sum([]byte(script))                                                                
    sum := hex.EncodeToString(sumRaw[:])
    println(sum)
}
```

### Expected result

```console
$ go run ./sample.go
aaf4c61ddcc5e8a2dabede0f3b482cd9aea9434d
```

### Got

```console
yaegi ./sample.go 
run: ./sample.go:11:28: cannot use type [20]uint8 as type []uint8
```

### Yaegi Version

v0.11.1

### Additional Notes

_No response_

mvertes:
Fixes #1328.

### DIFF
--- a/interp/type.go
+++ b/interp/type.go
@@ -3,7 +3,6 @@ package interp
 import (
 	"fmt"
 	"go/constant"
-	"log"
 	"path/filepath"
 	"reflect"
 	"strconv"
@@ -859,7 +858,6 @@ func nodeType2(interp *Interpreter, sc *scope, n *node, seen []*node) (t *itype,
 		case ptrT:
 			t = t.val
 		case valueT:
-			log.Println(n.cfgErrorf("nodeType sliceExpr"), t.rtype, t.rtype.Kind())
 			t = valueTOf(reflect.SliceOf(t.rtype.Elem()), withScope(sc))
 		}
 		if t.cat == arrayT {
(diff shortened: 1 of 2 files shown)


interp/type.go
```diff
@@ -854,8 +854,14 @@ func nodeType2(interp *Interpreter, sc *scope, n *node, seen []*node) (t *itype,
...
+		switch t.cat {
+		case ptrT:
 			t = t.val
+		case valueT:
+			switch t.rtype.Kind() {
+			case reflect.Array, reflect.Ptr:
+				t = valueTOf(reflect.SliceOf(t.rtype.Elem()), withScope(sc))
+			}
 		}
 		if t.cat == arrayT {
```
mpl, line 866: this should be moved as a case within the newly introduced switch L857, no?

interp/type.go
```diff
@@ -854,6 +854,13 @@ func nodeType2(interp *Interpreter, sc *scope, n *node, seen []*node) (t *itype,
 		if err != nil {
 			return nil, err
 		}
+
+		if t.cat == valueT {
+			switch t.rtype.Kind() {
+			case reflect.Array, reflect.Ptr:
+				t = valueTOf(reflect.SliceOf(t.rtype.Elem()), withScope(sc))
+			}
+		}
```
mpl, line 863:
```suggestion
			break
		}
```

mpl (approved):

interp/type.go
mpl: nevermind, I was confused by the succesion of switch and if. both indeed need to be done.

PR #1359, opened 2022-03-22
interp: fix unit testing for go1.18

mvertes:
Some tests are not passing when using go1.18, due to a change of
content in error messages compared to go1.17. We simply skip them
while we support go1.17. It concerns a small number of tests
regarding error detection.


ldez (approved):
LGTM

PR #1367, opened 2022-03-28
interp: fix handling of redeclaration in multi-assign expression

### LINKED ISSUE #1365: implicit return value not properly assigned

```go
package main

func genInt() (int, error) { return 3, nil }
    
func getInt() (value int) {
    value, err := genInt()
    if err != nil {
        panic(err)
    }
    return
}   
    
func main() {
    println(getInt())
}
```

### Expected result

```console
$ go run ./sample.go
3
```

### Got

```console
$ yaegi ./sample.go
0
```

### Yaegi Version

0.11.2

### Additional Notes

This issue was found when using github.com/spyzhov/ajson

mvertes:
In a statement like `a, b := f()` if `a` was previously declared,
and has the same type, its symbol must be reused, a new symbol must not override its
previous value. This is now fixed.

Fixes #1365.

### DIFF
--- a/_test/issue-1365.go
+++ b/_test/issue-1365.go
@@ -1,0 +1,18 @@
+package main
+
+func genInt() (int, error) { return 3, nil }
+
+func getInt() (value int) {
+	value, err := genInt()
+	if err != nil {
+		panic(err)
+	}
+	return
+}
+
+func main() {
+	println(getInt())
+}
+
+// Output:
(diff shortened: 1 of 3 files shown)


mpl (approved):
All good, but reminder for myself to ask you on discord: why is the fix needed only when the variable was declared as a return parameter?

PR #1380, opened 2022-04-12
interp: fix import of binary type symbols in current scope

### LINKED ISSUE #1360: dot import causes undefined type error on Slice or Array literal

```go
package main

import . "net"

func main() {
  // v := make(IP, 0) // works
  v := IP{} // <-  yaegi_test.go:19: 6:7: undefined type 
  println(v)
}
```

### Expected result

```console
<nil>
```

### Got

```console
yaegi_test.go:19: 6:7: undefined type
```

### Yaegi Version

v0.11.2

### Additional Notes

_No response_

mvertes:
Fixes #1360.

### DIFF
--- a/_test/issue-1360.go
+++ b/_test/issue-1360.go
@@ -1,0 +1,14 @@
+package main
+
+import (
+	"fmt"
+	. "net"
+)
+
+func main() {
+	v := IP{}
+	fmt.Println(v)
+}
+
+// Output:
+// <nil>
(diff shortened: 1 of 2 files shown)


mpl (approved):

mpl:
Note: we might have to check that we don't have the same problem for the import of a source package in "." , for a symbol of kind type. But that's for another issue/PR.

PR #1384, opened 2022-04-15
interp: expose fset to fix CompileAST issue

### LINKED ISSUE #1383: CompileAST cannot use imported packages

```go
func TestIssueX(t *testing.T) {
	const src = `
			package main

			func main() {
				fmt.Println("Hello")
			}
		`

	fset := token.NewFileSet()
	err := i.Use(stdlib.Symbols)
	if err != nil {
		t.Fatal(err)
	}
	_, err = i.Eval(`import "fmt"`)
	if err != nil {
		t.Fatal(err)
	}

	ast, err := parser.ParseFile(fset, "_.go", src, parser.DeclarationErrors)
	if err != nil {
		t.Fatal(err)
	}
	prog, err := i.CompileAST(ast)
	if err != nil {
		t.Fatal(err)
	}
	_, err = i.Execute(prog)
	if err != nil {
		t.Fatal(err)
	}
}
```

### Expected result

```console
$ go test
=== RUN   TestIssueX
Hello
--- PASS: TestIssueX (0.00s)
PASS
ok      github.com/traefik/yaegi/interp 0.007s
```

### Got

```console
$ go test
=== RUN   TestIssueX
    /home/firelizzard/src/yaegi/interp/interp_eval_test.go:1781: 1:1: undefined: fmt 17
--- FAIL: TestIssueX (0.00s)
FAIL
FAIL    github.com/traefik/yaegi/interp 0.007s
```

### Yaegi Version

5665c9a4108e5f878c0d1d8d1e1913eb3c314154

### Additional Notes

_No response_

firelizzard18:
The interpreter has its own internal fileset and expects all code to have been parsed using that fileset. If a user creates a fileset, calls `go/parser.Parse*`, then passes the result to `interp.CompileAST`, strange things can happen.

The solutions I can see are:

1. Expose the fileset so the user can use it when parsing source.
2. Add the fileset as an option (input to New) so that the user can tell the interpreter to use a specific fileset.
3. Pass the fileset into `CompileAST`

There are two ways to implement option 3. One is to add a field to nodes and update every reference to `interp.fset` to use `node.fset`. The other is to add a parameter to every function that references `interp.fset` or calls a function that does. Both of these are significant changes and involve an extra pointer for every node or most function calls.

Options 1 and 2 are easy. Option 2 involves adding an option so I went with option 1. I can imagine situations where option 2 could be necessary, but I can open another issue/PR if and when I need that.

Fixes #1383

### DIFF
--- a/interp/compile_test.go
+++ b/interp/compile_test.go
@@ -10,7 +10,8 @@ import (
 )
 
 func TestCompileAST(t *testing.T) {
-	file, err := parser.ParseFile(token.NewFileSet(), "_.go", `
+	i := New(Options{})
+	file, err := parser.ParseFile(i.FileSet(), "_.go", `
 		package main
 
 		import "fmt"
@@ -61,7 +62,6 @@ func TestCompileAST(t *testing.T) {
 		{desc: "expr", node: dFunc.Body.List[0]},
 	}
 
-	i := New(Options{})
 	_ = i.Use(stdlib.Symbols)
 
 	for _, c := range cases {
--- a/interp/interp_eval_test.go
+++ b/interp/interp_eval_test.go
@@ -5,6 +5,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
+	"go/parser"
 	"io"
 	"log"
 	"net/http"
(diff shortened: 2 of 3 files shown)


mvertes (approved):
LGTM

PR #1387, opened 2022-04-16
interp: add a function to get globals

### LINKED ISSUE #1385: Add call to expose variables declared with Eval
### Proposal

I want a function that lists all of the global variables in `package main`. Specifically, I want a list of variables declared with `Eval("var x = 1")` and `Eval("y := 1")`.

```go
// Globals returns a map of global variables and constants in the main package.
func (*Interpreter) Globals() map[string]reflect.Value
```

### Background

I have developed [an extension](https://gitlab.com/ethan.reesor/vscode-notebooks/go-playbooks) that implements Yaegi-based notebooks for Visual Studio Code. I want to add a variables view, like the Jupyter extension has for their notebooks.

### Workarounds

I don't think there are any. I was going to show a horrible hack using the debugging interface, but I don't think that exposes global variables. The Symbols function only returns exported values.

firelizzard18:
Closes #1385. Add `func (*Interpreter) Globals() map[string]reflect.Value` which returns a map of the global variables and constants declared in `package main`.

### DIFF
--- a/interp/interp_test.go
+++ b/interp/interp_test.go
@@ -1,9 +1,12 @@
 package interp
 
 import (
+	"go/constant"
 	"log"
 	"reflect"
 	"testing"
+
+	"github.com/traefik/yaegi/stdlib"
 )
 
 func init() { log.SetFlags(log.Lshortfile) }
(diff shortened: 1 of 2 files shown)


mvertes (approved):
LGTM

PR #1392, opened 2022-04-28
interp: fix the behaviour of goto, continue and break

### LINKED ISSUE #1354: Unreachable `goto` affects program behavior

```go
package main

import (
	"fmt"
	"regexp/syntax"
	"strings"
	"unicode/utf8"
)

const FuzzRegexp = "0"

var (
	_ = syntax.IsWordChar
	_ = strings.Index
)

type modeTypeFuzz uint8

const (
	modeMatchFuzz modeTypeFuzz = iota
	modeFirstFuzz
	modeLongestFuzz
)

// Fuzz implements the regular expression
// "0"
// with flags 212.
type Fuzz struct{}
type stateFuzz struct {
	c   [2]int
	i   int
	pc  int
	cnt int
}

// FindString returns the first leftmost match.
func (e Fuzz) FindString(r string) (matches [1]string, pos int, ok bool) {
	var bt [0]stateFuzz // static storage for backtracking state
	matches, pos, ok = e.doString(r, modeFirstFuzz, bt[:0])
	return
}

// FindLongestString returns the leftmost-longest match.
func (e Fuzz) FindLongestString(r string) (matches [1]string, pos int, ok bool) {
	var bt [0]stateFuzz // static storage for backtracking state
	matches, pos, ok = e.doString(r, modeLongestFuzz, bt[:0])
	return
}

func (e Fuzz) doString(s string, m modeTypeFuzz, bt []stateFuzz) (matches [1]string, pos int, ok bool) {
	var pmatches [1 * 2]int
	pmatches, ok = e.do(s, m, bt)
	pos = pmatches[0]

	for i := range matches {
		if pmatches[i*2] < 0 {
			continue
		}
		matches[i] = s[pmatches[i*2]:pmatches[i*2+1]]
	}

	return
}

func (e Fuzz) do(r string, m modeTypeFuzz, bt []stateFuzz) ([2]int, bool) {
	si := 0 // starting byte index
restart:
	bt = bt[:0]         // fast reset dynamic backtracking state
	c := [2]int{-1, -1} // captures
	var bc [2]int       // captures for the longest match so far
	matched := false    // succesful match flag
	i := si             // current byte index
	// fast prefix search "0"
	if idx := strings.IndexByte(r[si:], '0'); idx >= 0 {
		i += idx // prefix found, skip to it
		si = i
		c[0] = i   // start of match
		i += 1     // prefix length
		goto inst2 // instruction after prefix

	}
	i += len(r[si:]) // no prefix found, skip to the end of the rune slice

	c[0] = i   // start of match
	goto inst1 // initial instruction

	// inst0 unreacheable

	goto unreachable
	goto inst1
inst1: // string "0" -> 2
	if i >= 0 && len(r) >= i {
		if rs := r[i:]; len(rs) >= 1 && rs[:1] == "0" {
			i += 1
			goto inst2
		}
	}
	goto inst1_fail
	goto unreachable
	goto inst1_fail
inst1_fail:
	goto fail

	goto unreachable
	goto inst2
inst2: // match
	c[1] = i // end of match
	goto match

	goto unreachable
	goto fail
fail:
	{
		if i <= len(r) && len(bt) > 0 {
			switch bt[len(bt)-1].pc {
			default:
				panic(bt[len(bt)-1].pc)
			}
		}
		if matched {
			return bc, true
		}
		if len(r) > si {
			i = si
			cr, sz := rune(r[i]), 1
			if cr >= utf8.RuneSelf {
				cr, sz = utf8.DecodeRuneInString(r[i:])
			}

			si += sz
			_ = cr
			goto restart
		}
		return bc, false
	}

	goto unreachable
	goto match
match:
	if !matched || c[1]-c[0] > bc[1]-bc[0] {
		if m == modeMatchFuzz || m == modeFirstFuzz {
			return c, true
		}
		bc = c
		matched = true
	}
	goto fail

	goto unreachable
unreachable:
	panic("unreachable")
}

func FindString(s string) ([]string, int, bool) {
	m, p, f := Fuzz{}.FindString(s)
	return m[:], p, f
}

func main() {
	m, p, f := FindString("1101")
	fmt.Println(m, p, f)
}
```

### Expected result

```console
[0] 2 true
```

### Got

```console
[] 0 false
```

### Yaegi Version

14acf618af23471a62a1e7450b61b12eb5412d45

### Additional Notes

If the `goto inst2` immediately before the `inst2:` label is removed the problem disappears, even though that `goto` is unreachable and therefore never executed.

Sorry for the verbosity of the sample, the code is automatically generated.

mvertes:
The logic of goto was false due to mixing break label and goto
label, despite being opposite. In case of break, jump to node (the
exit point) instead of node.start. Also always define label symbols
before their use is parsed.

Fixes #1354.

### DIFF
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -220,20 +220,23 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
 			}
 
 		case breakStmt, continueStmt, gotoStmt:
-			if len(n.child) > 0 {
-				// Handle labeled statements.
-				label := n.child[0].ident
-				if sym, _, ok := sc.lookup(label); ok {
-					if sym.kind != labelSym {
-						err = n.child[0].cfgErrorf("label %s not defined", label)
-						break
-					}
-					sym.from = append(sym.from, n)
-					n.sym = sym
-				} else {
-					n.sym = &symbol{kind: labelSym, from: []*node{n}, index: -1}
-					sc.sym[label] = n.sym
+			if len(n.child) == 0 {
+				break
+			}
+			// Handle labeled statements.
+			label := n.child[0].ident
+			if sym, _, ok := sc.lookup(label); ok {
+				if sym.kind != labelSym {
+					err = n.child[0].cfgErrorf("label %s not defined", label)
+					break
 				}
+				n.sym = sym
+			} else {
+				n.sym = &symbol{kind: labelSym, index: -1}
+				sc.sym[label] = n.sym
+			}
+			if n.kind == gotoStmt {
+				n.sym.from = append(n.sym.from, n) // To allow forward goto statements.
 			}
 
 		case caseClause:
@@ -862,9 +865,7 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
 				err = n.cfgErrorf("invalid break label %s", n.child[0].ident)
 				break
 			}
-			for _, c := range n.sym.from {
-				c.tnext = n.sym.node
-			}
+			n.tnext = n.sym.node
 
 		case continueStmt:
 			if len(n.child) == 0 {
@@ -875,17 +876,15 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
 				err = n.cfgErrorf("invalid continue label %s", n.child[0].ident)
 				break
 			}
-			for _, c := range n.sym.from {
-				c.tnext = n.sym.node.child[1].lastChild().start
-			}
+			n.tnext = n.sym.node.child[1].lastChild().start
 
 		case gotoStmt:
 			if n.sym.node == nil {
+				// It can be only due to a forward goto, to be resolved at labeledStmt.
+				// Invalid goto labels are catched at AST parsing.
 				break
 			}
-			for _, c := range n.sym.from {
-				c.tnext = n.sym.node.start
-			}
+			n.tnext = n.sym.node.start
 
 		case labeledStmt:
 			wireChild(n)
@@ -893,9 +892,7 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
 				n.start = n.child[1].start
 			}
 			for _, c := range n.sym.from {
-				if c.kind == gotoStmt {
-					c.tnext = n.start
-				}
+				c.tnext = n.start // Resolve forward goto.
 			}
 
 		case callExpr:
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -202,9 +202,22 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
 					}
 				}
 			}
+
 			n.findex = -1
 			n.val = nil
 			sc = sc.pushBloc()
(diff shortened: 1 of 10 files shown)


interp/cfg.go
```diff
@@ -853,28 +854,49 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
...
-			} else {
+			if len(n.child) == 0 {
 				n.tnext = sc.loopRestart
+				break
+			}
+			if !n.hasAnc(n.sym.node) {
+				err = n.cfgErrorf("invalid continue label %s", n.child[0].ident)
+				break
+			}
+			for _, c := range n.sym.from {
```
mpl, line 878: Same remark as above.

interp/cfg.go
```diff
@@ -853,28 +854,49 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
...
+				c.tnext = n.sym.node.child[1].lastChild().start
 			}
 
 		case gotoStmt:
-			gotoLabel(n.sym)
+			if n.sym.node == nil {
+				break
+			}
+			for _, c := range n.sym.from {
```
mpl, line 886: same remark as in break and continue.

interp/cfg.go
```diff
@@ -853,28 +854,49 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
...
 
 		case labeledStmt:
 			wireChild(n)
 			if len(n.child) > 1 {
 				n.start = n.child[1].start
 			}
-			gotoLabel(n.sym)
+			for _, c := range n.sym.from {
+				if c.kind == gotoStmt {
+					c.tnext = n.start
```
mpl, line 897: If we are not in a forward case, it seems useless to do that, since the code in the goto case (L887) will overwrite c.tnext.

If we are in the forward case, we are overwriting c.tnext here. Is that what we want to do?

interp/cfg.go
```diff
@@ -853,28 +854,49 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
...
-			} else {
+			if len(n.child) == 0 {
 				n.tnext = sc.loop
+				break
+			}
+			if !n.hasAnc(n.sym.node) {
+				err = n.cfgErrorf("invalid break label %s", n.child[0].ident)
+				break
+			}
+			for _, c := range n.sym.from {
```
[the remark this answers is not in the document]
mvertes: True. Addressed in the next commit.

interp/cfg.go
```diff
@@ -853,28 +854,49 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
...
+				err = n.cfgErrorf("invalid continue label %s", n.child[0].ident)
+				break
+			}
+			for _, c := range n.sym.from {
+				c.tnext = n.sym.node.child[1].lastChild().start
 			}
 
 		case gotoStmt:
-			gotoLabel(n.sym)
+			if n.sym.node == nil {
```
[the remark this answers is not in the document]
mvertes: Syntax errors or invalid goto label are properly catched during AST parse. n.sym.node == nil is only possible in a case of a forward goto.

mpl (approved):

PR #1393, opened 2022-05-03
interp: improve handling of empty interface values

### LINKED ISSUE #1355: Interface with unexported method fails type assertion

```go
package main

import (
	"fmt"

	"github.com/traefik/yaegi/_test/p2"
)

func main() {
	var v *p2.T
	var i interface{}

	i = v
	if _, ok := i.(p2.I); ok {
		fmt.Printf("ok\n")
	} else {
		fmt.Printf("not ok\n")
	}
}
```

### Expected result

```console
$ go run ./sample.go
ok
```

### Got

```console
$ yaegi run ./sample.go
not ok
```

### Yaegi Version

14acf618af23471a62a1e7450b61b12eb5412d45

### Additional Notes

github.com/traefik/yaegi/_test/p2/p2.go is
```go
package p2

type I interface {
	isI()
}

type T struct{}

func (t *T) isI() {}
```

This is similar to #1025.

mvertes:
At variable, function parameter, slice, map or field element assign,
if the destination type is an empty interface, the value was never
wrapped into a valueInterface (to preserve type mutability in case
of re-assign). Now we wrap it in a valueInterface if the source
type has a non empty set of methods, to allow a future use as a non
empty interface.

There are still corner cases, but it extends notably the support
of interfaces within the interpreter.

Fixes #1355.

### DIFF
--- a/_test/issue-1355.go
+++ b/_test/issue-1355.go
@@ -1,0 +1,22 @@
+package main
+
+import "github.com/traefik/yaegi/_test/p2"
+
+func f(i interface{}) {
+	_, ok := i.(p2.I)
+	println("ok:", ok)
+}
+
+func main() {
+	var v *p2.T
+	var i interface{}
+
+	i = v
+	_, ok := i.(p2.I)
+	println("ok:", ok)
+	f(v)
+}
+
+// Output:
+// ok: true
+// ok: true
--- a/_test/p2/p2.go
+++ b/_test/p2/p2.go
@@ -1,0 +1,9 @@
+package p2
+
+type I interface {
+	isI()
+}
+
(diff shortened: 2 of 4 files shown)


mpl (commented):
To make the tests pass, and fix #1355 , I think we only need the changes in call and in genDestValue. But should I trust you, and assume we need the other changes for later anyway?

mvertes:
Yes, the issue is testing only function call, but we have to cover consistently (at least for maintainability) the other possible cases where an interpreter non empty interface value can be assigned to an empty interface value. I agree that we may miss unit tests for those cases.

mpl (approved):

PR #1405, opened 2022-05-24
interp: fix redeclaration of an interface variable

### LINKED ISSUE #1404: interface variable redefinition in a block

```go
package main

type I interface {
    inI()
}

type T struct {
    name string
}

func (t *T) inI() {}

func main() {
    var i I = &T{name: "foo"}

    if i, ok := i.(*T); ok {
        println(i.name)
    }   
}
```

### Expected result

```console
$ go run ./sample.go
foo
```

### Got

```console
$ yaegi ./sample.go 
run: ./sample.go:17:11: undefined selector: name
```

### Yaegi Version

v0.12.0

### Additional Notes

_No response_

mvertes:
Fixes #1404.


interp/cfg.go
```diff
@@ -2135,7 +2135,7 @@ func compDefineX(sc *scope, n *node) error {
 	for i, t := range types {
 		var index int
 		id := n.child[i].ident
-		if sym, level, ok := sc.lookup(id); ok && level == n.child[i].level && sym.kind == varSym && sym.typ.equals(t) {
+		if sym, level, ok := sc.lookup(id); ok && level == n.child[i].level && sym.kind == varSym && sym.typ.id() == t.id() {
```
mpl, line 2138: as an optimization, maybe we should keep the sym.typ.equals(t) check before doing the sym.typ.id() == t.id() check (instead of completely replacing it)? If the sym.typ.equals(t) is fast(er), and sometimes is enough to return false, then it would allow us to evaluate that the condition is false, without having to do the sym.typ.id() == t.id() one.
But I haven't benchmarked this at all, so I don't know.

PR #1406, opened 2022-05-24
interp: avoid panic when defining a label in incremental parsing mode

### LINKED ISSUE #982: stack overflow due to infinite recursion (such as when evaluating "a:a")
The following program `bug.go`, which generates and evaluates random 5 character sequences, triggers a fatal error. I have included a general case to test those random sequences, and a minimum case that crashes on a specific sequence.

**My code (general case):**
```go
package main

import (
    "context"
    "fmt"
    "math/rand"
    "time"

)

func main() {
    for {
        code := RandString(5)
        fmt.Println(code)
        //<-time.After(5 * time.Millisecond)
        _, _ = Run(code)
    }
}

func Run(code string) (ret interface{}, ok bool) {
    ok = true // until proven otherwise

    i.Use(stdlib.Symbols)

    timeout, cancel := context.WithTimeout(context.Background(), 100 * time.Millisecond)
    defer cancel()
    defer func() {
        if r := recover(); r != nil {
            fmt.Println("recovered from ", r, " with code ", code)
        }
    }()
    value, err := i.EvalWithContext(timeout, code)
    if err != nil {
        ok = false
    }
    if value.Kind() != 0 {
        ret = value.Interface()
    }
    return
}

// Util
var random = rand.New(rand.NewSource(time.Now().UnixNano()))

// bug also ocurrs with charset of "a:" as in a letter and a colon
var charset = []rune(" abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789\n:=(){}[]\"'")

func RandString(n int) string {
    b := make([]rune, n)
    for i := range b {
        b[i] = charset[random.Intn(len(charset))]
    }
    return string(b)
}

```
**My code (minimum case):**
```go
package main

import (
    "context"
    "fmt"
    "time"

)

func main() {
    Run("a:a")
    for{} // block so the program doesn't exit
}

func Run(code string) (ret interface{}, ok bool) {
    ok = true // until proven otherwise

    i.Use(stdlib.Symbols)

    timeout, cancel := context.WithTimeout(context.Background(), 100 * time.Millisecond)
    defer cancel()
    defer func() {
        if r := recover(); r != nil {
            fmt.Println("recovered from ", r, " with code ", code)
        }
    }()
    value, err := i.EvalWithContext(timeout, code)
    if err != nil {
        ok = false
    }
    if value.Kind() != 0 {
        ret = value.Interface()
    }
    return
}
```

**Expected result (general case):**
```console
$ go run bug.go
Rl859
5NoSI
RpRUg
 aWFM
CKWZ7
j "5y
(lFab
TPrvJ
v{H0'

...lots of 5 character sequences...

QO'A 
4[N(V
lI40K
00O)Z
s=niL
{WHM2
fLIIc
7cETq
```

**Got (general case):**
```console
$ go run bug.go
xi5pU
]Ck6'

...lots more 5 character sequences...

3WaVW
]p4Z4
b7Gs) <-- not sure if thise was the sequence that triggered the error, because console may not have flushed
runtime: goroutine stack exceeds 1000000000-byte limit
runtime: sp=0xc020688350 stack=[0xc020688000, 0xc040688000]
fatal error: stack overflow

runtime stack:
runtime.throw(0xe75f7a, 0xe)
	/home/finnb/sdk/go1.15.2/src/runtime/panic.go:1116 +0x72
runtime.newstack()
	/home/finnb/sdk/go1.15.2/src/runtime/stack.go:1060 +0x78d
runtime.morestack()
	/home/finnb/sdk/go1.15.2/src/runtime/asm_amd64.s:449 +0x8f

goroutine 322 [running]:
github.com/traefik/yaegi/interp.genValue(0xc00021c900, 0x0)
	/home/finnb/go/pkg/mod/github.com/traefik/yaegi@v0.9.8/interp/value.go:150 +0x60f fp=0xc020688360 sp=0xc020688358 pc=0x6a72cf
github.com/traefik/yaegi/interp.genValue(0xc00021c900, 0x0)
	/home/finnb/go/pkg/mod/github.com/traefik/yaegi@v0.9.8/interp/value.go:175 +0x285 fp=0xc020688428 sp=0xc020688360 pc=0x6a6f45

...lots more of the same thing (value.go:175 +0x285)...

github.com/traefik/yaegi/interp.genValue(0xc00021c900, 0x0)
	/home/finnb/go/pkg/mod/github.com/traefik/yaegi@v0.9.8/interp/value.go:175 +0x285 fp=0xc02068d0b8 sp=0xc02068cff0 pc=0x6a6f45
...additional frames elided...
created by github.com/traefik/yaegi/interp.(*Interpreter).EvalWithContext
	/home/finnb/go/pkg/mod/github.com/traefik/yaegi@v0.9.8/interp/interp.go:599 +0x150

goroutine 1 [runnable]:
github.com/traefik/yaegi/interp.(*Interpreter).Use(0xc0003e4000, 0xc0002028d0)
	/home/finnb/go/pkg/mod/github.com/traefik/yaegi@v0.9.8/interp/interp.go:645 +0x14b
main.Run(0xc00020e2f0, 0x5, 0x0, 0x0, 0x1)
	/home/finnb/git/finnbear/genetic-algorithm/bug.go:27 +0xb7
main.main()
	/home/finnb/git/finnbear/genetic-algorithm/bug.go:18 +0xbe
exit status 2
```

mvertes:
In REPL mode, a panic (stack overflow) could be triggered by:

	$ yaegi
	> a:
	runtime: goroutine stack exceeds 1000000000-byte limit
	runtime: sp=0x14020760330 stack=[0x14020760000, 0x14040760000]
	fatal error: stack overflow
	[...]

This issue occurs in incremental parsing mode only, and not when the parser
is in file mode. We avoid it by being more defensive when generating
values.

Fixes #982.

### DIFF
--- a/interp/value.go
+++ b/interp/value.go
@@ -34,10 +34,10 @@ func valueGenerator(n *node, i int) func(*frame) reflect.Value {
 // because a cancellation prior to any evaluation result may leave
 // the frame's data empty.
 func valueOf(data []reflect.Value, i int) reflect.Value {
-	if i < len(data) {
-		return data[i]
+	if i < 0 || i >= len(data) {
+		return reflect.Value{}
 	}
-	return reflect.Value{}
+	return data[i]
 }
 
 func genValueBinMethodOnInterface(n *node, defaultGen func(*frame) reflect.Value) func(*frame) reflect.Value {
@@ -195,7 +195,7 @@ func genValue(n *node) func(*frame) reflect.Value {
 		}
 		if n.sym != nil {
 			i := n.sym.index
-			if i < 0 {
+			if i < 0 && n != n.sym.node {
 				return genValue(n.sym.node)
 			}
 			if n.sym.global {
--- a/interp/interp_eval_test.go
+++ b/interp/interp_eval_test.go
@@ -1375,7 +1375,7 @@ func testConcurrentComposite(t *testing.T, filePath string) {
 	}
 }
 
-func TestEvalScanner(t *testing.T) {
+func TestEvalREPL(t *testing.T) {
 	if testing.Short() {
 		return
 	}
@@ -1460,6 +1460,13 @@ func TestEvalScanner(t *testing.T) {
 			src: []string{
 				`type bar string`,
 				`func (b bar) foo() { println(3) }`,
+			},
+			errorLine: -1,
+		},
+		{
+			desc: "define a label",
+			src: []string{
+				`a:`,
 			},
 			errorLine: -1,
 		},


interp/value.go
```diff
@@ -195,7 +195,7 @@ func genValue(n *node) func(*frame) reflect.Value {
 		}
 		if n.sym != nil {
 			i := n.sym.index
-			if i < 0 {
+			if i < 0 && n != n.sym.node {
```
mpl, line 198: Is this enough to also protect us in case of indirect recursion?
for example:
n != n.sym.node, but n == n.sym.node.sym.node ?

mpl (approved):

PR #1407, opened 2022-05-25
interp: create interpreter interface value with new

### LINKED ISSUE #1123: Package import got reflect error
Go version `go1.16.4` on `windows`

```go
package main

import (
)

func main() {
	i := interp.New(interp.Options{
		GoPath: "C:\\Users\\rageman\\go",
	})

	i.Use(stdlib.Symbols)

	_, err := i.Eval(`import "fmt"`)
	if err != nil {
		panic(err)
	}

	if _, err = i.Eval(`import "github.com/google/go-querystring/query"`); err != nil {
		panic(err)
	}

	_, err = i.Eval(`fmt.Println("Hello Yaegi")`)
	if err != nil {
		panic(err)
	}
}

```

Expected result:
```console
$ go run ./sample.go
// output
// successful import
```

Got:
```console
$ go run ./sample.go
impPath:github.com/google/go-querystring/query
interp.scopes: map[string]*interp.scope{"github.com/google/go-querystring/query":(*interp.scope)(0xc0000c7280), "main":(*interp.scope)(0xc0000c6980)}
run:_new *interp.frame <- *reflect.rtype typ:&reflect.rtype{size:0x10, ptrdata:0x10, hash:0x18a057e7, tflag:0x2, align:0x8, fieldAlign:0x8, kind:0x14, equal:(func(unsafe.Pointer, unsafe.Pointer) bool)(0x854780), gcdata:(*uint8)(0x1356bc0), str:77664, ptrToThis:290688}
C:\Users\rageman\go\src\github.com\google\go-querystring\query\encode.go:226:10: panic
panic: reflect.Set: value of type *interface {} is not assignable to type *interp.valueInterface

goroutine 1 [running]:
main.main()
        C:/Users/rageman/go/bin/yaegi_try.go:21 +0x194
```

Also if the import path is don't have any go files like `github.com/google/go-querystring` (not `github.com/google/go-querystring/query`) it panic nil pointer dereference, i think it because `interp.scopes[importPath]` not exist, `interp.scopes: map[string]*interp.scope{"main":(*interp.scope)(0xc000146980)}`
https://github.com/traefik/yaegi/blob/451c75406813ea990296abb9f423ae87390f91d6/interp/src.go#L123-L130

mvertes:
In that case, the interface must be wrapped in an valueInterface
at creation.

With that fix, it is now possible to import
github.com/google/go-querystring/query. Not tested beyond that.

Fixes #1123.

### DIFF
--- a/interp/run.go
+++ b/interp/run.go
@@ -3303,11 +3303,20 @@ func _len(n *node) {
 
 func _new(n *node) {
 	next := getExec(n.tnext)
-	typ := n.child[1].typ.TypeOf()
+	t1 := n.child[1].typ
+	typ := t1.TypeOf()
 	dest := genValueOutput(n, reflect.PtrTo(typ))
 
+	if isInterfaceSrc(t1) && (!isEmptyInterface(t1) || len(t1.method) > 0) {
+		typ = zeroInterfaceValue().Type()
+	}
+
 	n.exec = func(f *frame) bltn {
-		dest(f).Set(reflect.New(typ))
+		v := reflect.New(typ)
+		if vi, ok := v.Interface().(*valueInterface); ok {
+			vi.node = n
+		}
+		dest(f).Set(v)
 		return next
 	}
 }
--- a/interp/cfg.go
+++ b/interp/cfg.go
@@ -1324,7 +1324,7 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
 				// retry with the filename, in case ident is a package name.
 				sym, level, found = sc.lookup(filepath.Join(n.ident, baseName))
 				if !found {
-					err = n.cfgErrorf("undefined: %s %d", n.ident, n.index)
+					err = n.cfgErrorf("undefined: %s", n.ident)
 					break
 				}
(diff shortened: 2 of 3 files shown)


mpl (approved):

interp/run.go
```diff
@@ -3303,11 +3303,20 @@ func _len(n *node) {
...
+	if isInterfaceSrc(t1) && (!isEmptyInterface(t1) || len(t1.method) > 0) {
+		typ = zeroInterfaceValue().Type()
+	}
+
 	n.exec = func(f *frame) bltn {
-		dest(f).Set(reflect.New(typ))
+		v := reflect.New(typ)
+		if vi, ok := v.Interface().(*valueInterface); ok {
+			vi.node = n
+		}
```
mpl, line 3318: I noticed that this does not seem to be needed for the tests. Do you want to keep it anyway?

mvertes: Maybe a test is missing, but I'm sure that this change is correct and covers more cases than before.

mpl: ok

PR #1409, opened 2022-05-25
interp: improve type checking for defined types

### LINKED ISSUE #1408: type checking for defined types

```go
package main

type ( 
    Number  = int32
    Number2 = Number
)   
    
func f(n Number2) { println(n) }
    
func main() {
    var n Number = 5
    f(n)
}
```

### Expected result

```console
$ go run ./sample.go
5
```

### Got

```console
$ yaegi ./sample.go
run: ./sample.go:12:4: cannot use type main.Number as type main.Number2
```

### Yaegi Version

v0.12.0

### Additional Notes

_No response_

mvertes:
Fixes #1408.

### DIFF
--- a/interp/typecheck.go
+++ b/interp/typecheck.go
@@ -193,7 +193,7 @@ func (check typecheck) comparison(n *node) error {
 	ok := false
 	switch n.action {
 	case aEqual, aNotEqual:
-		if !isInterface(t0) && !isInterface(t1) && t0.untyped == t1.untyped && t0.id() != t1.id() {
+		if !isInterface(t0) && !isInterface(t1) && !t0.isNil() && !t1.isNil() && t0.untyped == t1.untyped && t0.id() != t1.id() {
 			// Non interface types must be really equals.
 			return n.cfgErrorf("invalid operation: mismatched types %s and %s", t0.id(), t1.id())
 		}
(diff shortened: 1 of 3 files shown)


interp/typecheck.go
```diff
@@ -184,23 +184,27 @@ func (check typecheck) shift(n *node) error {
...
 	}
 
 	ok := false
 	switch n.action {
 	case aEqual, aNotEqual:
-		ok = c0.typ.comparable() && c1.typ.comparable() || c0.typ.isNil() && c1.typ.hasNil() || c1.typ.isNil() && c0.typ.hasNil()
+		if !isInterface(t0) && !isInterface(t1) && !t0.isNil() && !t1.isNil() && t0.untyped == t1.untyped && t0.id() != t1.id() {
+			// Non interface types must be really equals.
+			return n.cfgErrorf("invalid operation: mismatched types %s and %s", t0.id(), t1.id())
+		}
```
mpl, line 199: I do not understand this part. Also, when I moved it above L193 (because I don't understand why we only need for the equality comparison), it didn't break any of the tests. So maybe we need a little bit more tests?
Anyway, you can explain it to me later if you're in a hurry, so I'll approve nonetheless.

mpl (approved):

PR #1413, opened 2022-06-02
interp: improve the behaviour of interface{} function parameters

### LINKED ISSUE #435: Interface defined in interpreter not picked up by fmt.Println
The following program `sample.go` has wrong output:

```go
package main

import (
	"fmt"
)

type Foo int

func (Foo) String() string {
	return "foo"
}

func print(arg interface{}) {
	fmt.Println(arg)
}

func main() {
	var arg Foo
	var f = print
	f(arg)
}
```

Expected result:
```console
$ go run ./sample.go
foo
```

Got:
```console
$ yaegi ./sample.go
0
```

It appears that `fmt.Println` does not see that `Foo` declared in the interpreter implements `fmt.Stringer` and simply outputs the int value.

I met the same problem with my interpreter gomacro and I have no idea how to fix this - except by "cheating" (i.e. replacing `fmt.Println` with an interpreter-aware implementation)

mvertes:
We finally address a long standing limitation of the interpreter:
the capacity to generate the correct interface wrapper for an
anonymous interface{} function parameter of a binary function.

It allows for example fmt.Printf to invoke the String method
of an object defined within the interpreter, or json.Marshal
to invoke a textMarshaler method if it exists and if there is
no Marshaler method already defined for the passed interpreter
object.

To achieve that, we add a new mapType part of the "Used" symbols
to describe what not empty interfaces are expected and in which
priority order. This information can not be guessed and is found
in the related package documentation, then captured in stdlib/maptypes.go.

Then, at compile time and/or during execution, a lookup on mapTypes
is performed to allow the correct wrapper to be generated.

This change adds a new MapType type to the stdlib package.

Fixes #435.

### DIFF
--- a/interp/run.go
+++ b/interp/run.go
@@ -1479,6 +1479,17 @@ func callBin(n *node) {
 		}
 	}
 
+	// getMapType returns a reflect type suitable for interface wrapper for functions
+	// with some special processing in case of interface{} argument, i.e. fmt.Printf.
+	getMapType := func(typ *itype) reflect.Type {
+		for _, rt := range n.interp.mapTypes[c0.rval] {
+			if typ.implements(&itype{cat: valueT, rtype: rt}) {
+				return rt
+			}
+		}
+		return nil
+	}
+
 	// Determine if we should use `Call` or `CallSlice` on the function Value.
 	callFn := func(v reflect.Value, in []reflect.Value) []reflect.Value { return v.Call(in) }
 	if n.action == aCallSlice {
@@ -1486,13 +1497,6 @@ func callBin(n *node) {
 	}
 
 	for i, c := range child {
-		var defType reflect.Type
-		if variadic >= 0 && i+rcvrOffset >= variadic {
-			defType = funcType.In(variadic)
-		} else {
-			defType = funcType.In(rcvrOffset + i)
-		}
-
 		switch {
 		case isBinCall(c, c.scope):
 			// Handle nested function calls: pass returned values as arguments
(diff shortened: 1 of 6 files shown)


interp/run.go
```diff
@@ -1527,6 +1534,19 @@ func callBin(n *node) {
 				break
 			}
 
+			// defType is the target type for an eventual interface wrapper.
+			var defType reflect.Type
+			if variadic >= 0 && i+rcvrOffset >= variadic {
+				defType = funcType.In(variadic)
+			} else {
+				defType = funcType.In(rcvrOffset + i)
+			}
```
mpl, line 1543: as an optimization, this could be in an else after the if of 1544.

mpl (approved):

interp/run.go
mvertes: I tested this change and tests fails at _test/issue-1166.go because `defType` is now nil when entering the switch at run.go:1551 (was not before).

mpl: One way that works is:
```
			// defType is the target type for a potential interface wrapper.
			var defType reflect.Type
			if getMapType != nil {
				if rt := getMapType(c.typ); rt != nil {
					defType = rt
				}
			}
			if defType == nil {
				if variadic >= 0 && i+rcvrOffset >= variadic {
					defType = funcType.In(variadic)
				} else {
					defType = funcType.In(rcvrOffset + i)
				}
			}
```

I haven't found an elegant way to invert the first condition yet though (in order to remove the `if defType == nil {`).

stdlib/maptypes.go
```diff
@@ -0,0 +1,58 @@
...
+	"fmt"
+	"log"
+	"reflect"
+)
+
+func init() {
+	mt := []reflect.Type{
+		reflect.TypeOf((*fmt.Formatter)(nil)).Elem(),
+		reflect.TypeOf((*fmt.Stringer)(nil)).Elem(),
+	}
```
cosmos72, line 16: What about adding also `fmt.GoStringer` ?

See https://pkg.go.dev/fmt#GoStringer

mvertes: In fact, the solution here is incomplete for `Printf`, as the non empty interface to target also depends in this case from the the format verbs, as `%#v` which requires a `GoStringer` instead of a `Stringer` for `%v`. For now we ignore this, and thus not adding `GoStringer`.

Having a full solution here for `Printf` would probably require to rewrite it inside the interpreter to parse the verbs in format string. It's too much of an edge case for us, and the current solution seems a better compromise.

PR #1415, opened 2022-06-13
interp: catch mismatched types for other comparisons

mpl:
The check for mismatched types was already added recently for ==  and != comparisons.
This PR now adds it for other comparisons ( < , <=, > , >=).


mvertes (approved):
LGTM

PR #1420, opened 2022-06-25
interp: improve support of alias types

tttoad:
I expect the following code to be supported.
```go
type TT http.Header

func (t TT) Set(key, val string) {

}

func (t TT) Get(key string) string {

}
```
So, I pushed this PR. 
Do I need to add some test cases?  I don't see the relevant test files ....

### DIFF
--- a/_test/alias3/alias3.go
+++ b/_test/alias3/alias3.go
@@ -1,9 +1,31 @@
 package alias3
 
+import (
+	"fmt"
+	"net/http"
+)
+
 type T struct {
 	A string
 }
 
 func (t *T) Print() {
 	println(t.A)
+}
+
+type A http.Header
+
+func (a A) ForeachKey() error {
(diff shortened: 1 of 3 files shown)


tttoad:
@mvertes 
Sorry, I always thought I was solving issue-1411...
The problem is that traversing aliasT.
So, I added two test cases.
The case added in alias4.go still does not work.
I need to continue testing this.

mvertes:
@tttoad, I see how to complete your fix. Some other small parts need to be changed as well in `cfg.go` and `typecheck.go`, won't explain there. I'll send those when I'm back from my leave Wednesday.
What you have done is useful and correct, thanks again.

tttoad:
@mvertes You're welcome.

mvertes (approved):
Looks good to me

PR #1430, opened 2022-07-20
stdlib: support of go1.18 and go1.19, remove go1.16 and go1.17

mvertes:
In addition:
- extract commmand now skips exported generics in runtime wrappers
- interp_consistent_test.go fixed for go1.18 and go1.19
- move minimal version of go compiler to go1.18

Note that this version is incompatible with go1.17 and before due
to the major changes in the stdlib go parser.

To be merged once go1.19 is officially released (not before).


ldez (approved):
LGTM

PR #1436, opened 2022-08-05
interp: fix generic check on nil function

mvertes:
Related to issue https://github.com/traefik/traefik/issues/9231


darkweak (approved):

ldez (approved):
Tested inside plugin validation service.

PR #1443, opened 2022-08-23
interp: improve handling of composed interfaces wrappers

### LINKED ISSUE #1425: stdlib._net_http_ResponseWriter is not seen as an http.Hijacker

```go
package main

import (
	"bufio"
	"errors"
	"log"
	"net"
	"net/http"
	"net/http/httputil"
	"net/url"
)

type Middleware struct {
	next http.Handler
}

func (h *Middleware) ServeHTTP(rw http.ResponseWriter, r *http.Request) {
	wrw := &WrappedWriter{writer: rw}
	h.next.ServeHTTP(wrw, r)
}

func main() {
	// forwards to e.g. websocat -s 6060
	dest, err := url.Parse("http://localhost:6060")
	if err != nil {
		log.Fatal(err)
	}
	next := httputil.NewSingleHostReverseProxy(dest)
	http.Handle("/", &Middleware{next})
	log.Fatal(http.ListenAndServe(":8080", nil))
}

type WrappedWriter struct {
	writer http.ResponseWriter
}

func (w *WrappedWriter) Header() http.Header {
	return w.writer.Header()
}

func (w *WrappedWriter) Write(buf []byte) (int, error) {
	return w.writer.Write(buf)
}

func (w *WrappedWriter) WriteHeader(statusCode int) {
	w.writer.WriteHeader(statusCode)
}

func (w *WrappedWriter) Hijack() (net.Conn, *bufio.ReadWriter, error) {
	h, ok := w.writer.(http.Hijacker)
	if !ok {
		return nil, nil, errors.New("hijack not supported")
	}
	return h.Hijack()
}
```

### Expected result

```console
As a backend, run e.g. (from https://github.com/vi/websocat , there's a brew for it):
$ websocat -s 6060

Start the above wrapped reverse proxy:
$ go run ./sample.go

Issue a websocket request on the reverse proxy (which will trigger the hijacking):
$ websocat ws://127.0.0.1:8080/
hello
("hello" should be printed on the backend output)
```

### Got

```console
$ yaegi run ./sample.go

$ websocat ws://127.0.0.1:8080/
websocat: WebSocketError: Received unexpected status code (502 Bad Gateway)
websocat: error running

Error on the yaegi side:
reverseproxy.go:490: http: proxy error: can't switch protocols using non-Hijacker ResponseWriter type stdlib._net_http_ResponseWriter
```

### Yaegi Version

cb642c44bad588189feb00760b5f0ec4ff5cdada (devel)

### Additional Notes

Related traefik issue:

https://github.com/traefik/traefik/issues/8682

mvertes:
This change implements a workaround to better support composed
interfaces in yaegi and let the interpreter define objects which
implement multiple interfaces at once.

We use the existing MapTypes to store what possible composed interface
wrapper could be used for some interfaces. When generating an interface
wrapper, the wrapper with the highest number of implemented methods is
chosen.

This is still an imperfect solution but it improves the accuracy of
interpreter in some critical cases.

This workaround could be removed in future if/when golang/go#15924
is resolved.

Fixes #1425.

### DIFF
--- a/interp/run.go
+++ b/interp/run.go
@@ -1050,19 +1050,23 @@ func genInterfaceWrapper(n *node, typ reflect.Type) func(*frame) reflect.Value {
 			return value
 		}
 	}
-	mn := typ.NumMethod()
+
+	// Retrieve methods from the interface wrapper, which is a struct where all fields
+	// except the first define the methods to implement.
+	// The first character of the field name must always be ignored.
+	wrap := getWrapper(n, typ)
+	mn := wrap.NumField() - 1
 	names := make([]string, mn)
 	methods := make([]*node, mn)
 	indexes := make([][]int, mn)
 	for i := 0; i < mn; i++ {
-		names[i] = typ.Method(i).Name
+		names[i] = wrap.Field(i + 1).Name[1:]
 		methods[i], indexes[i] = n.typ.lookupMethod(names[i])
 		if methods[i] == nil && n.typ.cat != nilT {
 			// interpreted method not found, look for binary method, possibly embedded
 			_, indexes[i], _, _ = n.typ.lookupBinMethod(names[i])
 		}
 	}
-	wrap := n.interp.getWrapper(typ)
 
 	return func(f *frame) reflect.Value {
 		v := value(f)
--- a/interp/run.go
+++ b/interp/run.go
@@ -1053,7 +1053,8 @@ func genInterfaceWrapper(n *node, typ reflect.Type) func(*frame) reflect.Value {
 
 	// Retrieve methods from the interface wrapper, which is a struct where all fields
 	// except the first define the methods to implement.
-	// The first character of the field name must always be ignored.
+	// As the field name was generated with a prefixed first character (in order to avoid
+	// collisions with method names), this first character is ignored in comparisons.
 	wrap := getWrapper(n, typ)
 	mn := wrap.NumField() - 1
 	names := make([]string, mn)
--- a/interp/use.go
+++ b/interp/use.go
@@ -77,7 +77,8 @@ func getWrapper(n *node, t reflect.Type) reflect.Type {
 	lm := n.typ.methods()
 
 	// mapTypes may contain composed interfaces wrappers to test against, from
-	// most complex to simplest. Return the first match, or the default.
+	// most complex to simplest (guaranteed by construction of mapTypes). Find the
+	// first for which the interpreter type has all the methods.
 	for _, rt := range n.interp.mapTypes[w] {
 		match := true
 		for i := 1; i < rt.NumField(); i++ {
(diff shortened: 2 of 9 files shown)


interp/run.go
```diff
@@ -1050,19 +1050,23 @@ func genInterfaceWrapper(n *node, typ reflect.Type) func(*frame) reflect.Value {
 			return value
 		}
 	}
-	mn := typ.NumMethod()
+
+	// Retrieve methods from the interface wrapper, which is a struct where all fields
+	// except the first define the methods to implement.
+	// The first character of the field name must always be ignored.
```
mpl, line 1056:
```suggestion
	// As the field name was generated with a prefixed first character (in order to avoid
	// collisions with method names), this first character is ignored in comparisons.
```

interp/use.go
```diff
@@ -0,0 +1,249 @@
...
+func getWrapper(n *node, t reflect.Type) reflect.Type {
+	p, ok := n.interp.binPkg[t.PkgPath()]
+	if !ok {
+		return nil
+	}
+	w := p["_"+t.Name()]
+	lm := n.typ.methods()
+
+	// mapTypes may contain composed interfaces wrappers to test against, from
+	// most complex to simplest. Return the first match, or the default.
```
mpl, line 80:
```suggestion
	// most complex to simplest (guaranteed by construction of mapTypes). Find the
	// first for which the interpreter type has all the methods.
```

interp/use.go
```diff
@@ -0,0 +1,249 @@
...
+			if _, ok := lm[rt.Field(i).Name[1:]]; !ok {
+				match = false
+				break
+			}
+		}
+		if match {
+			return rt
+		}
+	}
+
```
mpl, line 94:
```suggestion

	// Otherwise return the direct "non-composed" interface.
```

interp/use.go
```diff
@@ -0,0 +1,249 @@
...
+		m[k] = v
+		if importPath != "" {
+			return m
+		}
+	}
+
+	return m
+}
+
+// getWrapper returns the wrapper type of the corresponding interface, or nil if not found.
```
mpl, line 70:
```suggestion
// getWrapper returns the wrapper type of the corresponding interface, trying
// first the composed ones, or nil if not found.
```

stdlib/wrapper-composed.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (w _ioWriterReadFrom) Write(p []byte) (n int, err error) {
+	return w.WWrite(p)
+}
+
+func (w _ioWriterReadFrom) ReadFrom(r io.Reader) (n int64, err error) {
+	return w.WReadFrom(r)
+}
+
```
mpl, line 75:
```suggestion

// Each MapType value (each slice) must be sorted by complexity, i.e. by number
// of interface methods.
```

mpl (changes requested):

mpl (approved):

PR #1449, opened 2022-08-30
interp: fix redeclarations containing a blank variable

### LINKED ISSUE #1434: variable scope bug

```go
package main

import "fmt"

func main() {
	s := make([]map[string]string, 0)
	m := make(map[string]string)
	m["m1"] = "m1"
	m["m2"] = "m2"
	s = append(s, m)
	tmpStr := "start"
	fmt.Printf("%s\n", tmpStr)
	for _, v := range s {
		tmpStr, _ := v["m1"]
		fmt.Printf("%s\n",tmpStr)
	}
	fmt.Printf("%s\n", tmpStr)
}
```

### Expected result

```console
start
m1
start
```

### Got

```console
start
m1
m1
```

### Yaegi Version

master

### Additional Notes

_No response_

mvertes:
In [short variable declarations](https://go.dev/ref/spec#Short_variable_declarations),
The reuse of existing symbols is possible only if a new variable is defined,
otherwise a new symbol must be created, which was not the case in the issue.

Search for new symbols and correctly ignore blank variables.

Fixes #1434.


mpl (commented):
I do not think the fix is correct yet. Consider this very simple variation of the example issue:

```
package main

func main() {
	s := make([]map[string]string, 0)
	m := make(map[string]string)
	m["m1"] = "m1"
	m["m2"] = "m2"
	s = append(s, m)
	tmpStr := "start"
	println(tmpStr)
	for _, v := range s {
		tmpStr, ok := v["m1"]
		println(tmpStr)
		println(ok)
	}
	println(tmpStr)
}

// Output:
// start
// m1
// true
// start
```

mpl:
Also, maybe there's a confusion going on here. The example from the issue is _not_ a redeclaration, since a redeclaration is only in the same block.


this is a redeclaration:

```
	tmpStr := "start"
	println(tmpStr)
	tmpStr, ok := s[0]["m1"]
	println(tmpStr)
	println(ok)
```

but this is not:

```
	tmpStr := "start"
	println(tmpStr)
	for _, v := range s {
		tmpStr, ok := v["m1"]
		println(tmpStr)
		println(ok)
	}
	println(tmpStr)

```

which is also why, this is not allowed:

```
	tmpStr := "start"
	println(tmpStr)
	tmpStr, _ := s[0]["m1"]
	println(tmpStr)

```

whereas, this is allowed:
```
	tmpStr := "start"
	println(tmpStr)
	for _, v := range s {
		tmpStr, _ := v["m1"]
		println(tmpStr)
	}
	println(tmpStr)

```

interp/cfg.go
```diff
@@ -221,6 +221,12 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
 				sc.sym[label] = sym
 				c.sym = sym
 			}
+			// If block is the body of a function, get declared variables in current scope.
+			if n.anc != nil && n.anc.kind == funcDecl {
```
mpl, line 225:
```suggestion
			// This is done in order to add the func signature symbols into sc.sym,
			// as we will need them in post-processing.
			if n.anc != nil && n.anc.kind == funcDecl {
```

interp/cfg.go
```diff
@@ -2288,11 +2294,36 @@ func compDefineX(sc *scope, n *node) error {
...
+			hasNewSymbol = true
+		}
+	}
+
 	for i, t := range types {
 		var index int
 		id := n.child[i].ident
-		if sym, level, ok := sc.lookup(id); ok && level == n.child[i].level && sym.kind == varSym && sym.typ.id() == t.id() {
-			// Reuse symbol in case of a variable redeclaration with the same type.
+		// A variable can be redeclared if at least one none blank another variable is created.
```
mpl, line 2321:
```suggestion
		// A variable can be redeclared if at least one other not blank variable is created.
```

mpl (approved):

PR #1467, opened 2022-10-03
interp: fix type assertion for wrapped empty interface

mvertes:
Although empty interfaces are usually not wrapped, for compatibility with the runtime, we may have to wrap them sometime into `valueInterface` type.

It allows to preserve interpreter type metadata for interface values exchanged with the runtime. It is necessary to resolve methods and receivers in the absence of reflect support.

During type assertions on empty interfaces, we now handle a possible valueInterface and dereference the original value to pursue the type assertion.

In the same change, we have improved the format of some panic messages at runtime to give location of offending source at interpreter level.

This change will allow to fix traefik/traefik#9362.

### DIFF
--- a/_test/cli7.go
+++ b/_test/cli7.go
@@ -1,0 +1,57 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"log"
+	"net/http"
+	"net/http/httptest"
+)
+
+type T struct {
+	http.ResponseWriter
+}
+
+type mw1 struct {
+	next http.Handler
+}
+
+var obj = map[string]interface{}{}
+
+func (m *mw1) ServeHTTP(rw http.ResponseWriter, rq *http.Request) {
+	t := &T{
+		ResponseWriter: rw,
+	}
+	x := t.Header()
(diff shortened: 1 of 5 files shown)


_test/cli7.go
```diff
@@ -0,0 +1,57 @@
...
+	"net/http"
+	"net/http/httptest"
+)
+
+type T struct {
+	http.ResponseWriter
+}
+
+type mw1 struct {
+	next http.Handler
```
mpl, line 16: I think this field can be removed as it is not used, and does not bring anything to the table
```suggestion
```

mpl (approved):

_test/cli7.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Println(string(body))
+}
+
+// Output:
+// Welcome to my website!map[] &{<nil>}
```
mpl, line 56:
```suggestion
// Welcome to my website!map[] &{}
```

PR #1468, opened 2022-10-03
interp: check that send operate on channel value

### LINKED ISSUE #1453: panic on assignment to math.NaN()

```go
package main

import (
	"fmt"

)


func main() {
	interpreter := interp.New(interp.Options{})
	interpreter.Use(stdlib.Symbols)
	fmt.Println(interpreter.Eval("import \"math\""))
	fmt.Println(interpreter.Eval("math.NaN() <- 4"))
}
```

### Expected result

```console
0xc0000b8420 <nil>
<invalid reflect.Value> runtime error: invalid memory address or nil pointer dereference
```

### Got

```console
a panic about a nil pointer dereference
```

### Yaegi Version

v0.14.2

### Additional Notes

_No response_

mvertes:
Not performing this check was leading to a panic at run-time. It now fails early with a compile error.

Fixes #1453.


interp/cfg.go
```diff
@@ -960,7 +960,14 @@ func (interp *Interpreter) cfg(root *node, sc *scope, importPath, pkgName string
 			}
 			wireChild(n)
 
-		case declStmt, exprStmt, sendStmt:
+		case sendStmt:
+			if !isChan(n.child[0].typ) {
+				err = n.cfgErrorf("cannot send to non-channel %s", n.child[0].typ.id())
```
mpl, line 965: Shouldn't we prefix the message with "invalid operation: ", to make it match the go compiler?

```
package main

import "math"

func main() {
	math.NaN() <- 4
}
```

```
% go run /Users/mpl/main.go 
# command-line-arguments
/Users/mpl/main.go:6:2: invalid operation: cannot send to non-channel math.NaN() (value of type float64)
% 
% yaegi run /Users/mpl/main.go 
run: /Users/mpl/main.go:6:2: cannot send to non-channel float64
```

mvertes: Yes, it makes sense

PR #1471, opened 2022-10-12
interp: retry type definition if an array size is incomplete

### LINKED ISSUE #1470: out of order constant used in array size definition

```go
package main

type T struct {
        num [tnum + 2]int
}

const tnum = 23

func main() {
        t := T{}
        println(len(t.num))
}
```

### Expected result

```console
$ go run ./sample.go
25
```

### Got

```console
$ yaegi ./sample.go
./sample.go:11:14: undefined selector: num
```

### Yaegi Version

v0.14.2

### Additional Notes

_No response_

mvertes:
In case of forward definition of a constant, a symbol may be undefined when attempting to compute the array size in type ananlysis. Just mark the type as incomplete instead of aborting directly.

Fixes #1470.


interp/type.go
```diff
@@ -483,7 +483,11 @@ func nodeType2(interp *Interpreter, sc *scope, n *node, seen []*node) (t *itype,
 			length = int(vInt(sym.rval))
 		default:
 			// Size is defined by a numeric constant expression.
-			if _, err = interp.cfg(c0, sc, sc.pkgID, sc.pkgName); err != nil {
+			if _, err := interp.cfg(c0, sc, sc.pkgID, sc.pkgName); err != nil {
+				if strings.Contains(err.Error(), " undefined: ") {
```
mpl, line 487: this is horrible, but ok :)
Do you think we could a little bit more precise, and try to make sure that the error ends with the name of the undefined symbol? ("num" in the example).
I'm not sure it's possible though, because I've tried printing c0.ident and it seems it was empty at this point.
I'll let you be the judge and approve the PR anyway.

mpl (approved):

interp/type.go
mvertes: We need an additional helper to format the offending name from the AST. In the simplest case it could be the node ident, but not always and I not fully clear about the complexity. I think it should be the subject of another PR, and that we can merge this one.

PR #1472, opened 2022-10-12
interp: fix the logic to skip source files based on OS or CPU arch

mvertes:
For example, on architecture GOARCH=amd64, a file named `foobar_amd64.go` would be skipped instead of being read and parsed. The function `skipFile` is fixed and missing tests are added.


interp/build.go
```diff
@@ -147,10 +147,22 @@ func skipFile(ctx *build.Context, p string, skipTest bool) bool {
...
+				return y != ctx.GOARCH
+			}
+			return false
+		case knownOs[x] && knownArch[y]:
+			return true
+		case knownArch[y] && y != ctx.GOARCH:
+			return true
+		default:
+			return false
+		}
```
mpl, line 163: we could probably do better, but I didn't want to restart from scratch, so I modified just enough compared to your version to make the tests pass.

mvertes: Thanks for adding the missing tests and fixes. I can easily read and understand the logic of this code, so it is good for me!

mpl (approved):

PR #1473, opened 2022-10-12
interp: fix case behavior for values converted to empty interface

### LINKED ISSUE #1465: interface conversion: interface {} is *string, not interp.valueInterface

```go
package main

import (
	"fmt"
)

func SomeFunc[T int | string](defaultValue T) T {
	switch v := any(&defaultValue).(type) {
	case *string:
		*v = *v + " abc"
	case *int:
		*v -= 234
	}
	return defaultValue
}

func main() {
	fmt.Println(SomeFunc("test"))
	fmt.Println(SomeFunc(1234))
}
```

### Expected result

```console
test abc
1000
```

### Got

```console
main.go:8:2: panic
main.go:18:2: panic
run: interface conversion: interface {} is *string, not interp.valueInterface
```

### Yaegi Version

0.14.2

### Additional Notes

I stumbled on this error while I was creating some bridge wrappers for some external generic methods. 

Could reduce it to this example, hope someone can help me solve this one. As i do have a hard time understanding the internals of the interpreter that doing the cast at some point

[/github.com/traefik/yaegi@v0.14.2/interp/run.go:2999](https://github.com/traefik/yaegi/blob/master/interp/run.go#L2999)
```
vi := v.Interface().(valueInterface)
```

mvertes:
Fixes #1465.


interp/run.go
```diff
@@ -3043,13 +3049,23 @@ func _case(n *node) {
 					}
 					return fnext
 				}
-				if v := val.Interface().(valueInterface).node; v != nil {
-					for _, typ := range types {
-						if v.typ.id() == typ.id() {
-							destValue(f).Set(val)
-							return tnext
+				if vi, ok := val.Interface().(valueInterface); ok {
+					if v := vi.node; v != nil {
```
mpl, line 3053: why do we not have here first the equivalent of 3004-3009 ? i.e. when vi.node == nil and typ.cat == nitT ?

mpl (approved):
The changes look ok, but please see my comment about the missing vi.node case when we are on the multiple types part of the code.

mvertes (commented):
Ok to me.

PR #1487, opened 2022-11-04
Expose package name of a compiled source

dennwc:
Exposing package name of the compiled source allows using it in `Eval`, as well as in logs and for other purposes.


mvertes (approved):
Looks good to me

PR #1492, opened 2022-11-18
Fix deadlock after parsing directory with no files

dennwc:


dennwc:
cc @mvertes

mvertes (approved):
looks good to me. Thank you for contributing and fixing this issue.

Issue #1496, opened 2022-12-21
nil and slice equality is incorrect


elee1766:
### The following program `sample.go` triggers an unexpected result

```go
package main

import "fmt"

func main() {
	a := []byte{} == nil
	b := nil == []byte{}

	fmt.Printf("ans: %v\n", a)
	fmt.Printf("ans: %v\n", b)
}
```

### Expected result

```console
λ go run ./test.go                                                                                                                             
ans: false
ans: false
```

### Got

```console
λ yaegi ./test.go                                                                                                                              
ans: false
ans: true
```

### Yaegi Version

eee72d1aae664bf6627ef955215d886dc7b105c0

### Additional Notes

[]byte{} == nil and nil == []byte{} should both be false

however, yaegi thinks that nil == []byte{}

PR #1500, opened 2023-01-01
[WIP] #1496 Slice / Nil Equality

elee1766:
hi!

this issue is sorta blocking for me so i thought i would try to fix it. 

im still learning the codebase and understanding how yaegi works, but I thought I would attempt to add a test in the style of other tests as a start.

please let me know if there is anything i need to change / run, or if anyone knows perhaps a good place to start for tackling this issue

### DIFF
--- a/_test/issue-1496.go
+++ b/_test/issue-1496.go
@@ -1,0 +1,19 @@
+package main
+
+func ByteEqualsNil() bool {
+	return []byte{} == nil
+}
+
+func NilEqualsByte() bool {
+	return nil == []byte{}
+}
+
+func main() {
+	a := ByteEqualsNil()
+	b := NilEqualsByte()
+
+	println(a == false, b == false, a == b)
+}
+
+// Output:
+// true true true


elee1766:
ok, so these two statements `[]byte{} == nil` and `nil == []byte{}` seem to be getting optimized somewhere before interpretation? 

i say this because the "equals" in "op.go" is not called

elee1766:
ok, maybe third time is the charm

i added ```fmt.Printf(`nodes <%s> <%s> %v\n`, c0.name(), n.child[1].name(), value(f).IsNil()) ```

to the frame function near run.go:3880. it seems that the difference materializes here. 

so this is consistent with the behavior. It just checks if the first? value is nil, and if it is, returns true? hence why (nil==[]byte{}) returns true

but that means that uh, nil equality is completely broken broken then right?

ok

i updated the test just now, it seems that if nil is the first element in a comparison, the expression always evaluates to true.

elee1766:
questions:

1. i dont actually know if this check is 100% correct?  i think it is more correct than the previous one. it matches go run now at least.

2. i dont know if "nil" is valid in any binary operator that isnt `aEqual`. if its just aEqual, this fix should be fine?

(i couldn't run tests on my local computer, im not sure why, maybe i just cant read)

elee1766:
#1496 is this how github works trying to tag the issue to this pr

im so sorry if someone is subscribed to this issue and im spamming your in box

mvertes (changes requested):
Thank you for this nice first submission. The fix you propose is globally correct, just a few details to change before we can merge this.

Thanks again.

elee1766:
awesome. i think i changed those things. let me know if there is anything else that needs to change.

mvertes (approved):
Looks good to me.

PR #1522, opened 2023-03-11
More safeguards when searching for vendor root.

dennwc:
With certain yaegi configuration on Windows, the loop in `previousRoot` can be infinite, because it fails to recognize driver letters as root.

This change adds a few more safeguards: checks path prefix earlier and checks if `filepath.Dir` produces an empty path.


mvertes (approved):
LGTM

PR #1524, opened 2023-03-14
interp: fix type assertion issues

### LINKED ISSUE #1514: Unexpected behavior checking Public PEM key with `github.com/golang-jwt/jwt` package

```go
package main

import (
	"github.com/golang-jwt/jwt/v4"
	"log"
)

func main() {
	publicKeyPem := `
-----BEGIN PUBLIC KEY-----
MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQC03uKR6amRLogLQMS7OqT+x08g
Am8X5ScmfuYD+JktiqN9TQLRB8qQaclnaj+OjNEBnmt6Ttf1xBIwSPn+lmjjxhoZ
8gS5rbks30bqasHBFSJYaMlkWunIvhef89uswbjyyYB1Qyazf+jpmfZh9KWOJp8S
pGIMWgMqdk7MMA03IwIDAQAB
-----END PUBLIC KEY-----`

	_, err := jwt.ParseRSAPublicKeyFromPEM([]byte(publicKeyPem))

	if err != nil {
		log.Printf("%v", err)
	}
}
```

### Expected result

```console
Empty output (public key is correct)
```

### Got

```console
2023/02/15 15:04:52 key is not a valid RSA public key
```

### Yaegi Version

v0.15.0

### Additional Notes

it seems that the error comes from https://github.com/golang-jwt/jwt/blob/9358574a7a1a2c8d644f22b6e8de627ba85c58d0/rsa_utils.go#L100

ssbeatty:
closes #1514

hi!
I had the same problem as #1514 and I wanted to fix it, I found
When asserting *crypto/rsa.PublicKey, using the typ attribute of node to get an nil rtype, resulting in the assertion result being nok

So I submitted this Pull Request, hope it will be merged

### DIFF
--- a/interp/run.go
+++ b/interp/run.go
@@ -290,7 +290,7 @@ func typeAssert(n *node, withResult, withOk bool) {
 
 	typ := c1.typ // type to assert or convert to
 	typID := typ.id()
-	rtype := typ.rtype // type to assert
+	rtype := typ.refType(nil) // type to assert
 	next := getExec(n.tnext)
 
 	switch {


mvertes (changes requested):
Nice catch! Thank you for this excellent first submission.

Can you also add your test sample as `_test/assert3.go` so we have a good non regression test for this ?

I changed it to avoid using `log.Println` which displays a different date at each run, not suitable for reproduction and added expected output comment so it can be checked by the test suite.

Thanks again

```go
package main

import "crypto/rsa"

func main() {
    var pKey interface{} = &rsa.PublicKey{}

    if _, ok := pKey.(*rsa.PublicKey); ok {
        println("ok")
    } else {
        println("nok")
    }
}

// Output:
// ok
```

mvertes (approved):

PR #1533, opened 2023-03-25
extract: escape ~ in package name

sentriz:
consider packages from git.sr.ht, the user namespace is prefixed with `~`

so running something like `yaegi extract git.sr.ht/~emersion/scfg`

was producing syntax errors with `~` in identifiers. and also `~` in filenames which worked but probably best not to have it there either

thanks!


ldez:
Hello, can you sign the CLA? Thank you.

sentriz:
:+1:

mvertes (approved):
Looks good to me. Thank you for this first contribution.

PR #1551, opened 2023-05-12
fix: re-import of identical package with `sym.typ.cat == binPkgT`

he11olx:
test case:
```
func main() {
	script1 := `package script
import "encoding/json"
var _ = json.Marshal
`
	script2 := `package script
import "encoding/json"
var _ = json.Marshal
`
	inter := interp.New(interp.Options{GoPath: build.Default.GOPATH})
	_ = inter.Use(stdlib.Symbols)

	inter.Eval(script1)
	_, err := inter.Eval(script2)
	fmt.Println(err)
	// 2:8: json/_.go redeclared in this block
}
```

Related to: https://github.com/traefik/yaegi/pull/1066/files


he11olx:
Looks like not a bug but a feature. But In REPL, executing two `import "fmt" ` will also report errors.

```bash
➜  yaegi
> import "fmt"
: 0x14000030120
> import "fmt"
1:21: fmt/_.go redeclared in this block
```

PR #1556, opened 2023-06-13
fix(src): use `errors.Is(err, fs.ErrNotExist)`

laushunyu:
I implemented fs.FS. When the file does not exist, `xerrors.Wrap(fs.ErrNotExist, "")` will be returned. However, `os.IsNotExist` cannot handle this situation. I found the following comment:
```
// IsNotExist returns a boolean indicating whether the error is known to
// report that a file or directory does not exist. It is satisfied by
// ErrNotExist as well as some syscall errors.
//
// This function predates errors.Is. It only supports errors returned by
// the os package. New code should use errors.Is(err, fs.ErrNotExist).
```

Therefore, we should use `errors.Is(err, fs.ErrNotExist)` instead.


ldez:
@laushunyu you have to use an email related to your GitHub account to create your commit.
After that you will have to sign the CLA.

laushunyu:
> @laushunyu you have to use an email related to your GitHub account to create your commit. After that you will have to sign the CLA.

email corrected😇.

mvertes (approved):
Looks good to me. Thank you for this first contribution!

PR #1559, opened 2023-06-14
stdlib: remove embed wrapper

### LINKED ISSUE #1558: interface assertion not work

```go
package main

import (
	"fmt"
)

func main() {
	interpreter := interp.New(interp.Options{})
	interpreter.Use(stdlib.Symbols)
	expected := ""
	fmt.Printf(expected)

	_, err := interpreter.Compile(`package main

import "fmt"

type A struct {
}

func (a A) String() string {
	return "String"
}

func (a A) GoString() string {
	return "GoString"
}

func GetGoString() (fmt.GoStringer, bool) {
	var a fmt.Stringer = A{}
	goStr, ok := a.(fmt.GoStringer)
	return goStr, ok
}
`)
	if err != nil {
		panic(err)
	}

	mainSymbols := interpreter.Symbols("main")["main"]
	SelfSrc := mainSymbols["GetGoString"].Interface().(func() (fmt.GoStringer, bool))
	_, ok := SelfSrc()
	fmt.Println(ok)
}
```

### Expected result

```console
# go run .
true
```

### Got

```console
# go run .
false
```

### Yaegi Version

v0.15.1

### Additional Notes

hi,
assertion from a interface to another interface is not working.

I haven't found the exact location in the source code where this bug occurs yet, but I can suggest some possible causes and solutions for the bug:
```
package main

import "fmt"

type A struct {
}

func (a A) String() string {
	return "String"
}

func (a A) GoString() string {
	return "GoString"
}

func main() {
	var a fmt.Stringer = A{}
	v := reflect.ValueOf(&a).Elem()
	target := reflect.TypeOf((*fmt.GoStringer)(nil)).Elem()

	fmt.Println(v.Type(), target)
	fmt.Println(v.CanConvert(target))

	// implementation of interface v
	fmt.Println(v.Elem().Type(), target)
	fmt.Println(v.Elem().CanConvert(target))
}
```
output:
```
fmt.Stringer fmt.GoStringer
false
main.A fmt.GoStringer
true
```
I think when trying to know whether `v.CanConvert()`, we should `Elem` the v if `v.Kind == reflect.Interface`.

mvertes:
Embedding files using `//go:embed` and `embed` packages can not be supported in yaegi, so it is better to not generate the wrapper to embed and have a graceful error in case of usage of `embed.FS` in the interpreter.
Also update README about unsuported directives.

Fixes #1558.


ldez (approved):
LGTM

PR #1586, opened 2023-08-31
interp: Record function names in panic

dennwc:
Currently, yaegi only records source positions when writing panic trace to stderr.

This change adds function names to the panic output.

Unfortunately, yaegi walks the trace in reverse order, comparing to Go runtime. This can be improved in the future.


mvertes (approved):
Looks good to me. Thank you for this contribution.

PR #1587, opened 2023-09-02
interp: Add wasip1 to known OS list

dennwc:
Adds `wasip1` to known OS list, introduced in golang/go#58141.

Without this change, `yaegi extract` may fail on Go 1.21 with the following message:
```
type-checking package "time" failed (<GOROOT>/src/time/zoneinfo_wasip1.go:8:5: platformZoneSources redeclared in this block)
```

### DIFF
--- a/interp/build.go
+++ b/interp/build.go
@@ -182,6 +182,7 @@ var knownOs = map[string]bool{
 	"openbsd":   true,
 	"plan9":     true,
 	"solaris":   true,
+	"wasip1":    true,
 	"windows":   true,
 }
 


dennwc:
@mvertes this one is small enough, should be fast to review.

mvertes:
This will be addressed through bringing the full support of Go1.21 which is in preparation, thanks.

dennwc:
@mvertes I would advocate for merging it earlier, since it prevents `yaegi extract` from working with Go 1.21 installation at all, even if the project is configured for Go 1.20.

mvertes (approved):
LGTM

Issue #1606, opened 2024-01-25
Assignment statement compile panic


chikaku:
### The following program `sample.go` triggers an unexpected result

```go
package main

import "github.com/traefik/yaegi/interp"

const code = `
package main

func main() {
	a, b, c := 1, 2
	_, _, _ = a, b, c
}
`

func main() {
	i := interp.New(interp.Options{})
	_, err := i.Compile(code)
	println("err:", err.Error())
}
```

### Expected result

```console
err: cannot assign 2 values to 3 variables
```

### Got

```console
panic: runtime error: index out of range [5] with length 5 [recovered]
        panic: 5:2: CFG post-order panic: runtime error: index out of range [5] with length 5

goroutine 1 [running]:
github.com/traefik/yaegi/interp.(*Interpreter).cfg.func2.1()
        /gopath/pkg/mod/github.com/traefik/yaegi@v0.15.1/interp/cfg.go:601 +0x74
panic({0x12b0620?, 0xc0000a8090?})
        /usr/local/go/src/runtime/panic.go:920 +0x270
github.com/traefik/yaegi/interp.(*Interpreter).cfg.func2(0xc0000eb040)
        /gopath/pkg/mod/github.com/traefik/yaegi@v0.15.1/interp/cfg.go:642 +0x9bc5
github.com/traefik/yaegi/interp.(*node).Walk(0xc0000eb040, 0xc0000f5cc8, 0xc0000f5d10)
```

### Yaegi Version

v0.15.1

PR #1607, opened 2024-01-25
interp: fix mismatch assign statement panic (#1606)

chikaku:
Follow by the [Spec](https://go.dev/ref/spec#Assignment_statements):

The number of operands on the left hand side must match the number of values. For instance, if f is a function returning two values `x, y = f()` assigns the first value to x and the second to y.

In the second form, the number of operands on the left must equal the number of expressions on the right, each of which must be single-valued, and the nth expression on the right is assigned to the nth operand on the left.

### DIFF
--- a/_test/assign19.go
+++ b/_test/assign19.go
@@ -1,0 +1,9 @@
+package main
+
+func main() {
+	a, b, c := 1, 2
+	_, _, _ = a, b, c
+}
+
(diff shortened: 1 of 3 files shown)


mvertes (approved):
Looks good to me.

Thank you for your contrbution and this bug fix.

PR #1621, opened 2024-04-02
fix: avoid memory leak in closure

### LINKED ISSUE #1618: significant memory leak when spawning a Goroutine

```go
package main

import (
	"fmt"
	"runtime"
	"sync"
)

func humanizeBytes(bytes uint64) string {
	const (
		_         = iota
		kB uint64 = 1 << (10 * iota)
		mB
		gB
		tB
		pB
	)

	switch {
	case bytes < kB:
		return fmt.Sprintf("%dB", bytes)
	case bytes < mB:
		return fmt.Sprintf("%.2fKB", float64(bytes)/float64(kB))
	case bytes < gB:
		return fmt.Sprintf("%.2fMB", float64(bytes)/float64(mB))
	case bytes < tB:
		return fmt.Sprintf("%.2fGB", float64(bytes)/float64(gB))
	case bytes < pB:
		return fmt.Sprintf("%.2fTB", float64(bytes)/float64(tB))
	default:
		return fmt.Sprintf("%dB", bytes)
	}
}

func main() {
	i := 0
	wg := sync.WaitGroup{}
	for {
		var m runtime.MemStats
		runtime.ReadMemStats(&m)
		fmt.Printf("#%d: alloc = %s, routines = %d, gc = %d\n", i, humanizeBytes(m.Alloc), runtime.NumGoroutine(), m.NumGC)

		wg.Add(1)
		go func() {
			wg.Done()
		}()
		wg.Wait()
		i = i + 1
	}
}
```

### Expected result

```console
[...]
#1295805: alloc = 1.79MB, routines = 1, gc = 21
#1295806: alloc = 1.79MB, routines = 1, gc = 21
#1295807: alloc = 1.79MB, routines = 1, gc = 21
[...]
```

### Got

```console
[...]
#1046468: alloc = 9.35GB, routines = 1, gc = 33
#1046469: alloc = 9.35GB, routines = 1, gc = 33
#1046470: alloc = 9.35GB, routines = 1, gc = 33
[...]
```

### Yaegi Version

any

### Additional Notes

This isn't a goroutine leak, as the goroutines count remains stable and it isn't a garbage collector contention as the garbage collector is in the same range of iterations.

This hints at a leak in the Yaegi wrapper code around the execution of goStmt (yaegi/interp/run.go) and I'm wondering if a reference is not kept to a frame preventing its garbage collection and leaking a frame per goroutine call (just a theory at this point, I'm 0% familiar with Yaegi).

It causes traefik OOM when running plugins on a machine with frequent goroutine spawns (ie: a ticker).

mvertes:
Simplify frame management. Remove node dependency on frame pointer.

Fixes #1618

### DIFF
--- a/interp/run.go
+++ b/interp/run.go
@@ -1289,11 +1289,13 @@ func call(n *node) {
 	}
 
 	n.exec = func(f *frame) bltn {
+		f.mutex.Lock()
 		bf := value(f)
 		def, ok := bf.Interface().(*node)
 		if ok {
 			bf = def.rval
 		}
+		f.mutex.Unlock()
 
 		// Call bin func if defined
 		if bf.IsValid() {
@@ -1915,11 +1917,18 @@ func getFunc(n *node) {
 
 			// Interpreter code execution.
 			runCfg(n.child[3].start, fr2, n, n)
+
+			f.mutex.Lock()
 			getFrame(f, l).data[i] = o
+			f.mutex.Unlock()
+
 			return fr2.data[:numRet]
 		})
 
+		f.mutex.Lock()
 		getFrame(f, l).data[i] = fct
+		f.mutex.Unlock()
+
 		return next
 	}
 }


ldez (approved):
LGTM

PR #1625, opened 2024-04-25
fix: don't panic in map range if assigned value is _

mvertes:
Ensure that the code generated for `for s, _ = range ...` is the same as `for s = range ...`.

Fixes 1622.


ldez (approved):
LGTM

PR #1633, opened 2024-05-16
adding support to 'plugin' package

Raiane-Dev:
for cases where we have package import errors, it is good to have the option to link a plugin that contains these libraries.


mvertes (changes requested):
Thank you for this first contribution! In fact, to add a new generated wrapper, you need to do the following:

- edit the file listing the wrappers to be generated, and add the missing `plugin` name to it, for example at  https://github.com/traefik/yaegi/blob/c32b3ecfb6d71a2c8d30333b90de543e5e0819da/stdlib/stdlib.go#L54
- from project root, run `make generate` with go version go1.22
- `make generate` again with go version go1.21

You should obtain new generated files `stdlib/go1_21_plugin.go` and `stdlib/go1_22_plugin.go`.

The pull request should contain patches to `stdlib/stdlib.go`, and the generated files.

Thanks.

PR #1644, opened 2024-07-08
Range over int

rcoreilly:
This adds support for the `for i := range n` range-over-int form of the for loop, introduced in go 1.22  https://github.com/golang/go/issues/61405

Includes tests for 3 forms, and everything passes.


rcoreilly:
This is now superseded by #1645.

Issue #1634, opened 2024-05-17
Interpreted closures mixed with "binary" functions don't capture values correctly


theclapp:
### The following test case `TestIssue1634` in `interp/interp_eval_test.go` triggers an unexpected result

```go
func TestIssue1634(t *testing.T) {
	TLogf := t.Logf

	type CPE struct {
		E string
	}

	type CP struct {
		Es []*CPE
	}

	type Child struct {
		I int
		E string
		F func() string
	}

	FL := func(children []*Child) []string {
		s := []string{}
		for _, child := range children {
			s = append(s, child.F())
		}
		return s
	}

	R := func(i int, eE string, f func() string) *Child {
		TLogf("inside R: %d, %s", i, eE)
		return &Child{
			I: i,
			E: eE,
			F: f,
		}
	}

	var CpL func(cp *CP) []string
	// Having this on a line by itself makes it easy to copy into the interpreted
	// code below.
	CpL = func(cp *CP) []string {
		var ch []*Child
		for i, e := range cp.Es {
			i, e := i, e
			ch = append(ch,
				R(i, e.E, func() string {
					TLogf("inside R's f: %d, %v", i, e.E)
					return e.E
				}),
			)
		}
		return FL(ch)
	}

	var Cp *CP
	initCp := func() {
		Cp = &CP{
			Es: []*CPE{
				{E: "foo"},
				{E: "bar"},
				{E: "baz"},
			},
		}
	}
	initCp()

	// Verify that this works in standard Go
	t.Logf("Standard Go:")
	s := CpL(Cp)
	if expected, got := []string{"foo", "bar", "baz"}, s; !reflect.DeepEqual(expected, got) {
		t.Fatalf("Standard Go: CpL: Expected %v, got %v", expected, got)
	}

	i := interp.New(interp.Options{})
	err := i.Use(stdlib.Symbols)
	if err != nil {
		t.Fatalf("i.Use: Expected nil, got %v", err)
	}
	i.Use(interp.Exports{
		"pkg/pkg": {
			// vars
			"Cp":    reflect.ValueOf(&Cp).Elem(),
			"CpL":   reflect.ValueOf(&CpL).Elem(),
			"TLogf": reflect.ValueOf(&TLogf).Elem(),

			// funcs
			"FL": reflect.ValueOf(FL),
			"R":  reflect.ValueOf(R),

			// types
			"CP":    reflect.ValueOf((*CP)(nil)),
[...]

PR #1646, opened 2024-07-10
fix for issue #1634 -- for passing a closure to exported Go function

rcoreilly:
This fixes issue #1634 

includes special case for defer function.

I could remove or significantly reduce the comment description, and just have that here for future reference:

// per #1634, if v is already a func, then don't re-wrap!  critically, the original wrapping
// clones the frame, whereas the one here (below) does _not_ clone the frame, so it doesn't
// generate the proper closure capture effects!
// this path is the same as genValueAsFunctionWrapper which is the path taken above if
// the value has an associated node, which happens when you do f := func() ..

### DIFF
--- a/interp/interp_issue_1634_test.go
+++ b/interp/interp_issue_1634_test.go
@@ -1,0 +1,58 @@
+package interp
+
+import (
+	"bytes"
+	"io"
+	"os"
+	"reflect"
+	"testing"
+)
+
+func TestExportClosureArg(t *testing.T) {
+	outExp := []byte("0\n1\n2\n")
+	// catch stdout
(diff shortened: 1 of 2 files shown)


mvertes (commented):
Thank you for fixing this bug.

Ok to me, once the linter error on test case is fixed.

mvertes (approved):
LGTM

mvertes:
@rcoreilly if it is ok for you, allowing 'edit by maintainers' on next pull requests would permit us to better automate rebase at merge (when no conflicts) and manage not up-to-date branches.

In the meantime, you have to update the branch yourself, so we can proceed.

Thanks

Issue #1601, opened 2023-12-21
Feature request: support generic package


ahuigo:
### Proposal

Support generic package.
Is it possible?

### Background

It would be nice to use this famous generic package 	"github.com/samber/lo"

### Workarounds

I find that `yaegi` doesn't support generic package.
The following code shows an error: `reflect lo.Assign failed: cannot infer K `
```
$ ./bin/extract github.com/samber/lo
$ cat go1_21_github.com_samber_lo.go
package pkg

import (
	"github.com/samber/lo"
	"reflect"
)

func init() {
	Symbols["github.com/samber/lo/lo"] = map[string]reflect.Value{
		"Assign": reflect.ValueOf(lo.Assign),
....
```

Issue #1457, opened 2022-09-13
Import third party dependencies which are implemented with golang generic type


nyumaple07:
### Proposal

Hi there, 

Can we have support for adding third party dependencies which are implemented using golang generic type to an interpreter?

### Background

Let's have an example here:
```
type Integer[T int32 | int64] struct {
     someAttribute T
}
```

When I tried to add it to an interpreter, I had to add an actual type(which is `Integer[int32]` or `Integer[int64]`). Using that way, I can not use anything like `Integer[int32]` or `Integer[int64]` inside my script.
```
interpreter.Use(
map[string]map[string]reflect.Value{
    "xx/xx": {
    "Integer": reflect.ValueOf((*validation.Integer[int32])(nil)),
    },
})
```

### Workarounds

One workaround is to rename those types.

type Integer32 Integer[int32]
type Integer64 Integer[int64]

```
interpreter.Use(
map[string]map[string]reflect.Value{
    "xx/xx": {
    "Integer32": reflect.ValueOf((*validation.Integer[int32])(nil)),
    "Integer64": reflect.ValueOf((*validation.Integer[int64])(nil)),
    },
})
```

PR #1647, opened 2024-07-12
import generic functions from extracted Go code as interpreted code

rcoreilly:
This PR fixes issues like #1457 #1601 by importing the source code for generic functions in extracted Go code, so that it can be interpreted in yaegi.

You need to add a `//yaegi:add` comment directive to functions to so import, as importing all generic functions might not be desirable or feasible.  In particular, given the need to interpret the generic code, it can not access any non-exported functions, types, or fields.

The extracted code is wrapped in a `GenericFunc` string type value as below:

```go
 	err := i.Use(Exports{
		"guthib.com/generic/generic": map[string]reflect.Value{
			"Hello": reflect.ValueOf(GenericFunc("func Hello[T any](v T) { println(v) }")),
		},
	})
```

This PR builds on #1642 which interacts with this code in extracting the function source.


rcoreilly:
This and all the other PRs from me should now be good in terms of passing tests and working in our "real world" codebase which uses a lot of generics.

dzzzc:
Excuse me, will this feature be merged into the main branch?

rcoreilly:
how about merging this one too -- then we can go back to using this official version instead of our fork :)

coyzeng:
nil type for function call: likely generic type. can not invoke generic function.

PR #1648, opened 2024-07-12
adds helpful print commands for tracing and debugging the interp system.

rcoreilly:
I found these basic functions and Stringer methods essential for tracking down issues fixed recently -- probably be of use to others and I couldn't find something else in the code that would provide similar utility, though I never tried the debugger.

### DIFF
--- a/interp/trace.go
+++ b/interp/trace.go
@@ -1,0 +1,114 @@
+package interp
+
+import (
+	"fmt"
+	"reflect"
+	"strings"
+)
+
+// set trace to true for debugging the cfg and other processes
+var trace = false
+
+func traceIndent(n *node) string {
+	return strings.Repeat("  ", n.depth())
+}
+
+// tracePrintln works like fmt.Println, with indenting by depth
+// and key info on given node.
+func tracePrintln(n *node, v ...any) {
+	if !trace {
+		return
+	}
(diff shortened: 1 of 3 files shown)


interp/trace.go
```diff
@@ -0,0 +1,118 @@
...
+
+// nodeAddr returns the pointer address of node, short version.
+func ptrAddr(v any) string {
+	p := fmt.Sprintf("%p", v)
+	return p[:2] + p[9:] // unique bits
+}
+
+// valString returns string rep of given value, showing underlying pointers etc.
+//
+//lint:ignore
```
mvertes, line 48:
```suggestion
//nolint:unused // debugging facility
```

mvertes (commented):
Thanks for those text debugging functions. Their use is simple and straightforward. Definitely useful.

Just for info, up to now my debugging practice was to generate and display graphviz dot graphs for the AST and CFG, using the methods in interp/dot.go, and run something like:

```sh
YAEGI_DOT_CMD='dotty -' YAEGI_AST_CMD=1 go run ./cmd/yaegi
```

rcoreilly:
Thanks for the fixes and the pointer to the graphviz options -- didn't notice those before.

mvertes (approved):
LGTM

mvertes:
@rcoreilly please update, so we can merge it. Thanks.

PR #1677, opened 2025-01-06
fix: ensure that untyped values are converted to the same type

secDre4mer:
This fixes an issue where two untyped values with different `itype` could not be operated on.

Minimal example to reproduce this:

```
package main

import "math"

func main() {
	println(math.MaxFloat64 - 1.0)
}
```

Result: `invalid operation: mismatched types untyped float64 and untyped float`, but should be `1.7976931348623157e+308`

### DIFF
--- a/interp/typecheck.go
+++ b/interp/typecheck.go
@@ -270,6 +270,7 @@ func (check typecheck) binaryExpr(n *node) error {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


mvertes (approved):
Looks good to me.
Thank you for this contribution.

PR #1679, opened 2025-01-07
fix: correctly resolve fields of aliased structs

secDre4mer:
Fields of types that are aliases are not always resolved correctly, especially if the aliased type is itself an alias.

A minimal example is:
```
package main

type original struct {
	Field string
}

func main() {
	type alias original
	type alias2 alias
	var a = &alias2{
		Field: "test",
	}
	println(a.Field)
}
```

There is already a method (`baseType`) for resolving the underlying struct type; but it isn't always used previous to this change.

### DIFF
--- a/_test/type34.go
+++ b/_test/type34.go
@@ -1,0 +1,17 @@
+package main
+
+type original struct {
+	Field string
+}
+
+func main() {
+	type alias original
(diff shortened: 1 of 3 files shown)


mvertes (approved):
Looks good to me

PR #1680, opened 2025-01-08
chore: bump golangci-lint and fix config

mvertes:
This fixes the CI lint step. No change in code except a small fix in test output.


mloiseleur (approved):

mloiseleur:
@mvertes It fails on an other test. I launched it again, same result.

```sh
Error:     interp_consistent_test.go:355: got "# command-line-arguments\n../_test/switch13.go:9:7: i is not a type\n", want: "9:7: i (variable of type interface{}) is not a type"
```

mvertes:
I will fix the test prior merging. This is due to compiler error format change in the last versions of Go.

PR #1684, opened 2025-02-17
Feature: Support Go Modules

gcottom:
Adds support for go modules. Based on code by @james-lawrence . I found it useful. Made few modifications to ignore the main package in modules that have a main package and a package that contains the actual source. All tests should pass now which was the concern with the original PR related to this request in 2021. 

The code this is based on: [this](https://github.com/james-lawrence/yaegi/tree/0.14.3-modules-enh)

How it works: modules must be in the location that you specify using the GoPath option in the interp.New options. It doesn't download new modules for you. That is something you can script into your program, but this change isn't the place for that.


sf8193:
is there any reason this hasn't been merged in?

PR #1685, opened 2025-02-26
fix: use path instead of filepath

secDre4mer:
Currently, yaegi uses the `filepath` package to join paths.

However, this has issues: YAEGI uses an `fs.FS` internally for source code access, and `fs` specifies that `/` separated paths must be used. On Unix systems, this obviously makes no difference; but on Windows, with a non-default FS, this causes errors (since YAEGI creates `\` separated paths, but the FS expects `/` separated paths).

Furthermore, generally speaking, Golang import paths and the like are always `/` separated anyway, regardless of OS.

To clear this up, this PR changes path handling to `/` separated paths. This has the advantage of being accepted by all OS (even Windows will accept `/` separated paths). Paths passed by the user to `CompilePath` or similar are changed to `/` separators, if necessary.


mvertes (approved):
Sorry for the delay in reviewing this. Thank you for fixing the support of windows. Looks good to me.

PR #1710, opened 2025-09-09
Support Partial Type Inference

nelsam:
When a function has multiple type parameters, it is possible to explicitly pass in some types but rely on type inference for others. For example, `func Foo[T, U any](v U) T { ... }` may be called with `Foo[string](12)`.

This implements support for partial type inference for both `indexExpr` and `indexListExpr` ast.


nelsam:
I've found a use case that this breaks; it seems like I need to put more work into it.

PR #1715, opened 2026-02-06
Fix nil pointer dereference when sending to binary channel type alias

pk910:
## Description

When sending a value to a channel that is a type alias defined in a binary package, yaegi panics with a nil pointer dereference.

This PR was created with AI assistance (Claude) as I hit this issue in a downstream project using yaegi. 
The fix and tests have been reviewed and verified for correctness and functionality by myself.

## Root Cause

In the `send` function (`interp/run.go`), the code directly accesses `c0.typ.val` to get the channel element type:

```go
value1 := genDestValue(c0.typ.val, c1)
```

For source-defined channels (category `chanT`), `.val` contains the element type. But for binary channel type aliases (category `valueT`), `.val` is nil - the element type is only available via reflection in `.rtype`.

## Fix

Use the existing `elem()` method which correctly handles both cases:

```go
value1 := genDestValue(c0.typ.elem(), c1)
```

The `elem()` method (in `type.go`) already handles this properly:

```go
func (t *itype) elem() *itype {
    if t.cat == valueT {
        return valueTOf(t.rtype.Elem())  // Binary types: create from reflect
    }
    return t.val  // Source types: use existing itype with all metadata
}
```

This is a one-line fix that fixes binary channel type aliases by using reflection to get the element type

## Test Results

| Test | Without Fix | With Fix |
|------|-------------|----------|
| `TestSendToSourceDefinedChannel` | PASS | PASS |
| `TestSendToSourceDefinedChannelTypeAlias` | PASS | PASS |
| `TestSendToBinaryChannelTypeAlias` | FAIL | PASS |

### mypkg/mypkg.go

```go
package mypkg

// IntChan is a channel type alias
type IntChan chan int

// NewIntChan creates a new IntChan
func NewIntChan() IntChan {
    return make(IntChan, 1)
}
```

### symbols.go

```go
package main

import (
    "reflect"
    "yaegi-repro/mypkg"
)

var Symbols = map[string]map[string]reflect.Value{}

func init() {
    Symbols["yaegi-repro/mypkg/mypkg"] = map[string]reflect.Value{
        "IntChan":    reflect.ValueOf((*mypkg.IntChan)(nil)),
        "NewIntChan": reflect.ValueOf(mypkg.NewIntChan),
    }
}
```

### main.go

```go
package main

import (
    "fmt"
    "log"
)

func main() {
    if err := i.Use(Symbols); err != nil {
        log.Fatal(err)
    }

    code := `
package main

import "yaegi-repro/mypkg"

func main() {
    ch := mypkg.NewIntChan()
    ch <- 42  // This line panics without the fix
    val := <-ch
    println("Received:", val)
}
`
    _, err := i.Eval(code)
    if err != nil {
        log.Fatal(err)
    }
    fmt.Println("Success!")
}
```

## Results

**Without fix:**
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x128 pc=0x...]

goroutine 1 [running]:
github.com/traefik/yaegi/interp.(*itype).refType(0x0?, ...)
    .../interp/type.go:2065 +0x7a
github.com/traefik/yaegi/interp.(*itype).TypeOf(...)
    .../interp/type.go:2223
github.com/traefik/yaegi/interp.genDestValue(0x0, ...)
    .../interp/value.go:157 +0x25
github.com/traefik/yaegi/interp.send(...)
    .../interp/run.go:3766 +0x...
```

**With fix:**
```
Received: 42
Success!
```

### DIFF
--- a/interp/interp_chan_test.go
+++ b/interp/interp_chan_test.go
@@ -19,7 +19,7 @@ func TestSendToBinaryChannelTypeAlias(t *testing.T) {
 	i := interp.New(interp.Options{})
 
 	err := i.Use(interp.Exports{
-		"test/test": {
+		"mypkg/mypkg": {
 			"IntChan":    reflect.ValueOf((*IntChan)(nil)),
 			"NewIntChan": reflect.ValueOf(NewIntChan),
 		},
@@ -31,10 +31,10 @@ func TestSendToBinaryChannelTypeAlias(t *testing.T) {
 	_, err = i.Eval(`
 package main
 
-import "test"
+import "mypkg"
 
 func main() {
-	ch := test.NewIntChan()
+	ch := mypkg.NewIntChan()
 	ch <- 42
 	val := <-ch
 	if val != 42 {
--- a/interp/interp_chan_test.go
+++ b/interp/interp_chan_test.go
@@ -1,0 +1,92 @@
+package interp_test
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/traefik/yaegi/interp"
+)
+
+// IntChan is a channel type alias for testing.
+type IntChan chan int
+
+// NewIntChan creates a new IntChan.
+func NewIntChan() IntChan {
+	return make(IntChan, 1)
+}
+
+func TestSendToBinaryChannelTypeAlias(t *testing.T) {
+	i := interp.New(interp.Options{})
+
+	err := i.Use(interp.Exports{
+		"test/test": {
+			"IntChan":    reflect.ValueOf((*IntChan)(nil)),
+			"NewIntChan": reflect.ValueOf(NewIntChan),
+		},
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	_, err = i.Eval(`
+package main
+
+import "test"
+
+func main() {
+	ch := test.NewIntChan()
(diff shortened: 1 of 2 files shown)


interp/interp_chan_test.go
```diff
@@ -0,0 +1,92 @@
...
+// NewIntChan creates a new IntChan.
+func NewIntChan() IntChan {
+	return make(IntChan, 1)
+}
+
+func TestSendToBinaryChannelTypeAlias(t *testing.T) {
+	i := interp.New(interp.Options{})
+
+	err := i.Use(interp.Exports{
+		"test/test": {
```
mvertes: Please do not use "test" for the binary package name as it can be confused with the standard library test package (for testing). Use something like "mypkg" instead.

mvertes (changes requested):
The fix is correct. The test case needs to be adjusted to avoid confusion with standard library test package.

pk910:
Heya @mvertes 
Thanks for your review :)
I've updated the test accordingly.

mvertes (approved):
Looks good to me

traefiker:
:no_entry_sign: the mergeable state is "blocked"

Issue #1674, opened 2024-12-19
min/max functions are undefined


lboynton:
### The following program `sample.go` triggers an unexpected result

```go
package main

import "fmt"

func main() {
	fmt.Println(max(1, 2))
}
```

### Expected result

```console
go run ./sample.go
2
```

### Got

```console
$ yaegi ./sample.go
run: sample.go:6:14: undefined: max
```

### Yaegi Version

v0.16.1

### Additional Notes

These functions should be part of Go 1.21: https://go.dev/ref/spec#Min_and_max

Spuxy:
@mvertes i would like to take and implement it, but is this project some kind of dead/unmaintained ? or is there any chance to accept my PR ? (obviously if the code will be valid)

PR #1721, opened 2026-05-05
added builtin function min() and max()

Spuxy:
Added min and max builtin functions as i discussed before, i took this https://github.com/traefik/yaegi/issues/1674 because of lack of this functionality.


andig:
Seems yaegi is no longer maintained?

PR #1729, opened 2026-07-27
fix: don't panic on a constant non-bool condition

Endika:
A non-bool condition in `if` or `for` is reported as an error when it is a variable, but
crashes when it is a constant:

```go
package main

func main() {
        if 1 {
        }
}
```

```console
$ yaegi run cond.go
panic: reflect: call of reflect.Value.Bool on int64 Value [recovered]
        panic: cond.go:4:2: CFG post-order panic: reflect: call of reflect.Value.Bool on int64 Value
```

Expected, and what `var i int; if i {}` already produces today: `4:5: non-bool used as if
condition`. String, float and typed constants panic the same way, and so do the `for` forms.
Since the panic escapes, it takes down a host program that embeds yaegi.

The check records the error but does not stop the case, so execution reaches the
compile-time-constant fast path and calls `rval.Bool()` on a value that is not a bool. A
non-constant condition has no valid `rval` and takes the `else` branch, which is why only the
constant case crashes. This is the part left uncovered when #438 was fixed.

The fix breaks out of the case as soon as the error is recorded, as the other `cfgErrorf` sites
in this switch already do, for all eight condition forms. Five `_test` files cover the plain
`if`, the `if/else` form, `if` with an init statement, the bare `for` and the three-clause
`for`.


Endika:
The Linting failure here isn't this PR — the job is broken for the whole repo.

GO_VERSION: stable now resolves to go1.26.5, while GOLANGCI_LINT_VERSION: v2.4.0 is a prebuilt binary built with go1.25. Its go/types can't type-check a go1.26 GOROOT, so make check panics before it lints anything:

go version go1.26.5 linux/amd64
...
golangci-lint run
panic: file requires newer Go version go1.26 (application built with go1.25) [recovered, repanicked]
go/types.(*Checker).handleBailout(0xc0183983c0, 0xc00715da20)
      go/types/check.go:467 +0x97

No diagnostic is emitted, and Build and Test and Checks code and generated code are skipped because they need linting. The ten Go ({stable,oldstable}, {ubuntu,macos,windows}-latest) jobs on this branch all pass, so nothing here fails to build.

It also predates this PR: the previous Main run (#1721, 2026-05-05) failed at Linting too, though its logs have since expired.

Every golangci-lint release from v2.9.0 on is built with Go 1.26 (GO_VERSION: '1.26' in their release.yml; v2.8.0 was still 1.25), so bumping GOLANGCI_LINT_VERSION — v2.12.2 is current — should get the job running again. One thing worth deciding while you're there: with GO_VERSION: stable the pair drifts apart on every Go release, so this will come back when stable moves to go1.27. Pinning the lint job to the Go version golangci-lint was built with avoids the next round.

Happy to open a separate CI PR for the bump if that's useful.

